PR #110, opened 2023-02-13
Carousel component

guerrap:
Implement a new `Carousel` component that will be used in the main frontend homepage for displaying the featured/latest campaigns, the templates, and the oracles in the erc20 template for the creation form.


Diogomartf:
I'm not much fan of the arrows on desktop.
Also they mess with scrolling. eg. When I'm scrolling down, and the carousel appears, it starts to scroll carousel without me engaging in it.

image

Do we really need them? Or scrolling as It was would be enough?

Diogomartf:
On mobile I like the scroll. It is not correctly aligned on my device.

image

image

Diogomartf:
> Hey @Diogomartf, ye the carousel is still a bit clunky, we still have to tweak it.

My bad, just saw the wip! 

either way, that's the feedback I have now.

PR #123, opened 2023-02-17
New light mode multi step card style and next button

luzzif:


packages/ui/src/components/navigation/multi-step-cards/next-button/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+    children,
+    ...rest
+}: NextButtonProps): ReactElement => {
+    return (
+        <div className="cui-absolute cui-bottom-0 cui-left-1/2 cui-transform -cui-translate-x-1/2 cui-translate-y-1/2">
+            <Button
+                {...rest}
+                size="small"
+                className={{
+                    // FIXME: find out why important is needed, fix
```
guerrap, line 18: About this we may need to look into this I think https://github.com/dcastil/tailwind-merge, but we'll see.

luzzif: Good suggestion; added [this](https://github.com/carrot-kpi/v1-monorepo/issues/124) issue to make us remember

PR #125, opened 2023-02-17
Add stepper component

guerrap:
Add a `Stepper` component to the UI lib.


Diogomartf:
Just a small improv suggestion:

It would be great when clicking on for eg step2 the orange line also changed on the story. Currently only the content changes.

image

packages/ui/src/components/navigation/stepper/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+        },
+    },
+});
+
+type StepperLayout = "vertical" | "horizontal";
+
+export interface StepperProps {
+    stepTitles: string[];
+    activeStep: number;
+    mostUpdatedStep: number;
```
Diogomartf, line 71: what does `mostUpdatedStep` means?

Diogomartf: Maybe `lastStepUpdated`  describes better.

guerrap: The `mostUpdatedStep` is the last completed step, it's used to set the boundary for the navigation; the user can navigate to the already completed steps.

Ye maybe `lastStepCompleted` is a bit more clear.

guerrap:
> Just a small improv suggestion:
> 
> It would be great when clicking on for eg step2 the orange line also changed on the story. Currently only the content changes.
> 
> image

Do you mean something like this?
[scrnli_2_17_2023_6-31-45 PM.webm](https://user-images.githubusercontent.com/9011637/219726059-50f175ba-fd1f-452b-ab3b-4b4348dac5b3.webm)

Diogomartf:
> 1.00
> > Just a small improv suggestion:
> > It would be great when clicking on for eg step2 the orange line also changed on the story. Currently only the content changes.
> > image
> 
> Do you mean something like this?

Yes! It wasn't working when I tested.

guerrap:
> > 1.00
> > > Just a small improv suggestion:
> > > It would be great when clicking on for eg step2 the orange line also changed on the story. Currently only the content changes.
> > > image
> > 
> > 
> > Do you mean something like this?
> 
> Yes! It wasn't working when I tested.

It wasn't implemented yet :sweat_smile:

packages/ui/src/components/navigation/stepper/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+            ],
+        },
+        active: {
+            true: ["cui-bg-orange"],
+            false: ["cui-bg-black"],
+        },
+    },
+});
+
+type StepperLayout = "vertical" | "horizontal";
```
luzzif, line 70: I think it would make sense for us to have an `automatic` layout type automatically switching from `vertical` to `horizontal` on < `md`

luzzif (changes requested):

luzzif:
Sers it doesn't make sense to me that the orange line also moves back when you click on a previous step. The orange line simply indicates where a user is in the form, and should always arrive to the most up to date step.

guerrap:
> Sers it doesn't make sense to me that the orange line also moves back when you click on a previous step. The orange line simply indicates where a user is in the form, and should always arrive to the most up to date step.

Maybe from the video is not really visible, but the last completed step is identified by a square with an orange border, while the currently active one is orange.

luzzif:
> > Sers it doesn't make sense to me that the orange line also moves back when you click on a previous step. The orange line simply indicates where a user is in the form, and should always arrive to the most up to date step.
> 
> Maybe from the video is not really visible, but the last completed step is identified by a square with an orange border, while the currently active one is orange. This way is clear where you are in the stepper and which is the most updated step.

Uh that's a nice compromise. Yeah it's not that noticeable from the video. Lesgo then

PR #126, opened 2023-02-17
Extract animated modal to its own component and improve styles

luzzif:


Diogomartf (dismissed):
Much better 👏

packages/frontend/src/components/fullscreen-modal/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+                orange: ["bg-orange"],
+                green: ["bg-green"],
+            },
+        },
+    }
+);
+
+interface FullscreenModalProps {
+    bgColor?: NavbarProps["bgColor"];
+    style: object;
```
guerrap, line 20: Maybe we can improve the typing here? The `object` could be replaced with `SpringValues` I think.

luzzif: Absolutely, addressed in 78d4a2ad972972ec0f1b83234b460c01cc4b4497

PR #127, opened 2023-02-17
feat: rewrite useLatestKPITokens hook with react query

Diogomartf:
Add react query and first replacement on useLatestKPITokens hook. 

This is the first hook, just to try out react query as see if it fits our workflow.

React query docs 👇 
https://tanstack.com/query/v4/docs/react/quick-start

The main benefits for us are easy catching and maintainable queries. It will allow us to remove a lot of repetitive code.


packages/frontend/src/bootstrap.tsx
```diff
@@ -45,14 +49,21 @@ export const Root = ({
...
             <ThemeUpdater />
-            <CarrotCoreProvider
-                supportedChains={supportedChains || standaloneSupportedChains}
-                providers={providers || standaloneProviders}
-                getConnectors={connectors || getStandaloneConnectors}
-                ipfsGatewayURL={ipfsGatewayURL}
-            >
-                <App customBaseURL={customBaseURL} templateId={templateId} />
-            </CarrotCoreProvider>
+            <QueryClientProvider client={queryClient}>
```
luzzif, line 52: Pretty sure the plan is to eventually also use React Query in the `@carrot-kpi/react` lib. That's why I think it's better to have the Query Provider there, inside the `CarrotCoreProvider`.

packages/frontend/src/hooks/useLatestKPITokens.ts
```diff
@@ -12,44 +13,46 @@ export function useLatestKPITokens(limit = 5): {
...
-    const [loading, setLoading] = useState(true);
-
-    useEffect(() => {
-        let cancelled = false;
-        async function fetchData(): Promise<void> {
-            if (!chain) return;
-            if (!cancelled) setLoading(true);
-            try {
-                const kpiTokensAmount = await Fetcher.fetchKPITokensAmount({
+    const { data: kpiTokensAmountData, isSuccess: isKPITokensAmountSuccess } =
```
luzzif, line 16: Isn't it better to have all in one `queryFn`? This is an atomic operation and I don't see where the single cached items would be used elsewhere.

luzzif (changes requested):

packages/frontend/src/bootstrap.tsx
Diogomartf: Yes, but then we have to move all the hooks there, otherwise it will not be possible to use it in places like `frontend/src/hooks/useLatestKPITokens.ts`.

packages/frontend/src/hooks/useLatestKPITokens.ts
Diogomartf: What do you mean by having all in on `queryFn`?

Diogomartf: This were 3 requests already (dependent on each other):
`Fetcher.fetchKPITokensAmount`
`Fetcher.fetchKPITokenAddresses`
`Fetcher.fetchKPITokens`
 
Just replaced all with react query, I think the code is simpler.

packages/frontend/src/bootstrap.tsx
luzzif: >Yes, but then we have to move all the hooks there, otherwise it will not be possible to use it in places like frontend/src/hooks/useLatestKPITokens.ts.

Wait that's not true. The `CarrotCoreProvider` wraps the frontend's `App` so we can use React query in both the React app and the frontend. That's the same principle we use for the wagmi client, which is shared between the React lib and the frontend. In order to do that it's important to add React Query to peer dependencies in the React lib and to the dependencies in the frontend.

packages/frontend/src/hooks/useLatestKPITokens.ts
luzzif: Yeah but it doesn't make sense for those 3 requests to have each a cache entry since they act as one single bigger request. That's why I'd just have one cache entry for the latest x KPI tokens, not 3 different entries one for the amount, one for the addresses and one for the tokens themselves.

packages/frontend/src/bootstrap.tsx
Diogomartf: Yeah, it makes sense it works on both, but I couldn't make it work. I'll try again.

packages/frontend/src/hooks/useLatestKPITokens.ts
Diogomartf: So you're suggesting to only add the last query `Fetcher.fetchKPITokens` to react query, and do the others as before?

luzzif: Nono, to add all of them. Basically just take the `fetchData` function that was used in the hook and pass it as is (with changes due to the fact that you can't use `setState`) to the `queryFn` to React Query.

Diogomartf: I don't think "useQuery" can be called inside a `useEffect`.

Diogomartf: but why it's bad to have the 3 queries with react query? what I'm I missing?

I think the code looks clear as well.

Diogomartf: updated on 59dc180017597dd9a9a1c4b89013aae1e566eab4

packages/frontend/src/bootstrap.tsx
Diogomartf: wasn't able to do this, I tried to add peer dependencies as it is with wagmi but it fails.

`React package`
image
**frontend package**
image
 
also added `QueryClientProvider` to react package and remove it from bootstrap.

image


fails with:

`No QueryClient set, use QueryClientProvider to set on
`

packages/frontend/src/hooks/useLatestKPITokens.ts
luzzif: > but why it's bad to have the 3 queries with react query? what I'm I missing?
> 
> Is less efficient by catching 3 times?

Yeah, pretty sure it's less efficient to have 3 responses cached when the 3 requests as a whole act like an atomic request overall.

luzzif:
Just done some changes. There were 2 versions of React query floating around and that's why it didn't work across the React library and frontend. I noticed it by looking at the lockfile directly, but pretty sure a `yarn list \@tanstack/react-query` would have highlighted the issue too. So I've fixed that + implemented localstorage cache for queries + slightly changed the `queryKey` of the hook. It doesn't contain `preferDecentralization` now. There's no need to refetch stuff if that setting changed, it's ok to do it for subsequent calls only.

Diogomartf:
> Just done some changes. There were 2 versions of React query floating around and that's why it didn't work across the React library and frontend. I noticed it by looking at the lockfile directly, but pretty sure a `yarn list \@tanstack/react-query` would have highlighted the issue too. So I've fixed that + implemented localstorage cache for queries + slightly changed the `queryKey` of the hook. It doesn't contain `preferDecentralization` now. There's no need to refetch stuff if that setting changed, it's ok to do it for subsequent calls only.

thanks for the help, do we want to use "offline first" network approach ?

packages/frontend/src/bootstrap.tsx
```diff
@@ -24,6 +24,18 @@ import {
...
 } from "./standalone-setup";
+import { QueryClient } from "@tanstack/react-query";
+
+const queryClient = new QueryClient({
+    defaultOptions: {
+        queries: {
+            cacheTime: 1_000 * 60 * 60 * 24, // 24 hours
+            networkMode: "offlineFirst",
+            refetchOnWindowFocus: false,
+            retry: false,
```
Diogomartf, line 35: In offline first mode it will run the `queryFn` once, but then pause retries. 

why do we need `retry` `false` here?

packages/react/src/providers/core-provider/index.tsx
```diff
@@ -63,14 +73,25 @@ export const CarrotCoreProvider = ({
...
+            key: "carrot-kpi-react-query-cache",
+            storage: window.localStorage,
+        }),
+    });
+
     if (!!ipfsGatewayURL) IPFSService.gateway = ipfsGatewayURL;
+
     return (
         <PreferencesProvider>
             <WagmiSetup
```
Diogomartf, line 90: doesn't make sense to me to pass the react query client as props to `WagmiSetup` when I read this I think they are related.

Can we just put the react query bellow `PreferencesProvider`?

luzzif: Ah sheesh, now that you say it you're right. Fixing

packages/frontend/src/bootstrap.tsx
luzzif: I actually just copied over the config from Wagmi. Thought they would use some sensible configuration themselves, so mostly took it. I'm not sure the retries feature is suuuuper useful to us, but I can enable it if needed.

PR #132, opened 2023-02-20
Implement first version of preferences menu

luzzif:


Diogomartf:
The buttons should have the same height here:

image

I think the connect button is too big (and not the settings one), can we make it smaller?

Diogomartf:
Toggling dark mode doesn't seems to work.
image

packages/frontend/src/components/ui/navbar/index.tsx
```diff
@@ -39,25 +42,25 @@ const navbarStyles = cva(
...
+// const navLinksStyles = cva(["flex"], {
+//     variants: {
+//         isOpen: {
+//             true: ["flex-col items-start space-y-8 relative"],
+//             false: [
+//                 "items-center space-x-8 left-1/2 absolute transform -translate-x-1/2 -translate-y-1/2",
+//             ],
+//         },
+//     },
+// });
```
Diogomartf, line 63: removing this code, breaks the open menu on mobile.

image

packages/ui/src/global.css
```diff
@@ -15,7 +15,7 @@
 }
 
 @layer components {
-    .cui-switch {
+    button.cui-switch {
```
Diogomartf, line 18: why this css class need to be tied to button?

luzzif:
> The buttons should have the same height here:
> 
> image
> 
> I think the connect button is too big (and not the settings one), can we make it smaller?

Addressed in 79e64986d90444e9c81793406b5044bbfe795693

luzzif:
> Toggling dark mode doesn't seems to work. image

Addressed in 554bd76624d23c315e8ae1789a14c3c10faac190

packages/ui/src/global.css
luzzif: Otherwise Tailwind's base CSS overrides the switch style.

Diogomartf:
We need more spacing on mobile between the _burguer_ icon and the _settings_ icon.
image

Diogomartf (approved):

packages/frontend/src/components/info-popover/index.tsx
```diff
@@ -0,0 +1,40 @@
+import { Popover } from "@carrot-kpi/ui";
+import React, { ReactNode, useCallback, useRef, useState } from "react";
+import { ReactComponent as Info } from "../../assets/info.svg";
+
+interface InfoPopoverProps {
+    children?: ReactNode;
+}
+
+export const InfoPopover = ({ children }: InfoPopoverProps) => {
```
guerrap, line 9: Maybe we can move this component to the `UI` lib, I think it can be useful if we need to add hints to the inputs in the creation form for example, wdyt?

guerrap (approved):

packages/frontend/src/components/info-popover/index.tsx
luzzif: I thought about that but I'd like to add some functionality before sending it to the UI lib. Namely, I'd like to turn this to a generalized component with `info`, `warning` and `error` variants with dedicated icons/colors and such. I'll add a `TODO` though.

PR #133, opened 2023-02-20
Minor UI components fixes

guerrap:
Minor tweaks for the `card` and `accordions` components.

### DIFF
--- a/packages/ui/src/components/surfaces/accordion/index.tsx
+++ b/packages/ui/src/components/surfaces/accordion/index.tsx
@@ -1,6 +1,15 @@
 import { cva } from "class-variance-authority";
-import React, { ReactElement, useCallback, useMemo, useState } from "react";
+import React, {
+    ReactElement,
+    ReactNode,
+    useCallback,
+    useMemo,
+    useState,
+} from "react";
+import { matchChildByType } from "../../../utils/components";
 import { AccordionContextProvider } from "./context";
+import { AccordionDetails } from "./details";
+import { AccordionSummary } from "./summary";
 
 export * from "./details";
 export * from "./summary";
@@ -23,7 +32,7 @@ export interface AccordionProps {
     onChange?: (event: React.MouseEvent, expanded: boolean) => void;
     expanded?: boolean;
     className?: { root?: string };
-    children: ReactElement[];
+    children: ReactNode[];
 }
 
 export const Accordion = ({
@@ -35,14 +44,22 @@ export const Accordion = ({
 }: AccordionProps): ReactElement => {
     const [internalExpanded, setInternalExpanded] = useState<boolean>(false);
 
+    const childSteps = useMemo(
+        () => React.Children.toArray(children),
+        [children]
+    );
     const isControlled = useMemo(() => expanded !== undefined, [expanded]);
     const isExpanded = useMemo(
         () => (isControlled ? !!expanded : internalExpanded),
         [isControlled, expanded, internalExpanded]
     );
 
-    const summaryChildren = children[0];
-    const detailsChildren = children[1];
+    const summaryChildren = childSteps.find((child) =>
+        matchChildByType(child, AccordionSummary)
+    );
+    const detailsChildren = childSteps.find((child) =>
+        matchChildByType(child, AccordionDetails)
+    );
 
     const handleOnClick = useCallback(
         (event: React.MouseEvent) => {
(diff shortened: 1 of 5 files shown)


Diogomartf (approved):

packages/ui/src/components/surfaces/accordion/index.tsx
```diff
@@ -23,7 +32,7 @@ export interface AccordionProps {
     onChange?: (event: React.MouseEvent, expanded: boolean) => void;
     expanded?: boolean;
     className?: { root?: string };
-    children: ReactElement[];
+    children: ReactNode[];
```
luzzif, line 35: Does it make sense to have an array here? I think `ReactNode` supports passing stuff like:

```
<div />
<div />
<div />
```

No?

packages/ui/src/components/surfaces/accordion/index.tsx
```diff
@@ -35,14 +44,22 @@ export const Accordion = ({
...
+    );
     const isControlled = useMemo(() => expanded !== undefined, [expanded]);
     const isExpanded = useMemo(
         () => (isControlled ? !!expanded : internalExpanded),
         [isControlled, expanded, internalExpanded]
     );
 
-    const summaryChildren = children[0];
-    const detailsChildren = children[1];
+    const summaryChildren = childSteps.find((child) =>
```
luzzif, line 57: We've used useMemo above and 2 loops here. Wondering if it makes sense to have a reduce on the children in the `useMemo` to just do one single loop and have both the accordion summary and details.

packages/ui/src/components/surfaces/card/index.tsx
```diff
@@ -41,11 +41,7 @@ export const Card = ({ className, children }: CardProps): ReactElement => {
 
     return (
         <div className={rootStyles({ className: className?.root })}>
-            {titleChildren && (
-                <div className="h-12 cui-flex cui-items-center cui-border-b cui-border-black dark:cui-border-white [&:last-of-type]:cui-border-none">
-                    {titleChildren}
-                </div>
-            )}
+            {titleChildren && titleChildren}
```
luzzif, line 44: Can't we just have `titleChildren`?

luzzif (changes requested):

packages/ui/src/components/surfaces/accordion/index.tsx
guerrap: I meant to change `ReactElement` to `ReactNode` only `Summary` and `Details`, my bad. I'll revert this change.

packages/ui/src/components/surfaces/accordion/index.tsx
guerrap: Do you mean something like this?
```tsx
const { summaryChildren, detailsChildren } = React.Children.toArray(
    children
).reduce<{
    summaryChildren: ReactNode;
    detailsChildren: ReactNode;
}>(
    (previous, current) => {
        if (matchChildByType(current, AccordionSummary)) {
            return {
                ...previous,
                summaryChildren: current,
            };
        }
        if (matchChildByType(current, AccordionDetails)) {
            return {
                ...previous,
                detailsChildren: current,
            };
        }

        return previous;
    },
    { summaryChildren: null, detailsChildren: null }
);
````

File: packages/ui/src/components/surfaces/accordion/index.tsx [hunk 4]
```diff
@@ -41,8 +50,30 @@ export const Accordion = ({
         [isControlled, expanded, internalExpanded]
     );
 
-    const summaryChildren = children[0];
-    const detailsChildren = children[1];
+    const { summaryChildren, detailsChildren } = React.Children.toArray(
```
luzzif, line 53: Spreading previous shouldn't be necessary in a reduce, we can just push. What about something like this?

```
const { summaryChildren, detailsChildren } = useMemo(() => {
        return React.Children.toArray(children).reduce(
            (
                accumulator: {
                    summaryChildren: ReactNode[];
                    detailsChildren: ReactNode[];
                },
                child
            ) => {
                if (matchChildByType(child, AccordionSummary))
                    accumulator.summaryChildren.push(child);
                else if (matchChildByType(child, AccordionDetails))
                    accumulator.detailsChildren.push(child);
                return accumulator;
            },
            { summaryChildren: [], detailsChildren: [] }
        );
    }, [children]);
```

luzzif (changes requested):

PR #136, opened 2023-02-21
Add button to tags input

guerrap:
Add a button to the tags input for a more intuitive ux (also on mobile it wasn't possible to add tags).

### DIFF
--- a/packages/ui/src/components/input/tags/index.tsx
+++ b/packages/ui/src/components/input/tags/index.tsx
@@ -1,13 +1,22 @@
 import { cva } from "class-variance-authority";
 import React, { forwardRef, useCallback, useState } from "react";
 import { ReactElement } from "react";
+import { Button } from "../button";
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/input/tags/index.tsx
```diff
@@ -1,13 +1,22 @@
...
+    "cui-flex-wrap",
+    "cui-gap-2",
+    "cui-mt-2",
+]);
+
+const inputWrapperStyles = cva(["cui-flex", "cui-gap-2"]);
 
 export type TagsInputProps = Omit<BaseInputProps<string[]>, "onChange"> & {
     onChange: (tags: string[]) => void;
+    addButtonLabel: string;
```
luzzif, line 19: The practice as of now has been to add a `message` object to components that need labels passed in.

luzzif (changes requested):

PR #138, opened 2023-02-22
Persist theme preferences

guerrap:
Persist the interface settings preferences on local storage.


packages/frontend/src/components/connect-wallet/popovers/preferences/index.tsx
```diff
@@ -40,7 +42,10 @@ export const PreferencesPopover = forwardRef<
             <div className="flex justify-between gap-20 items-center">
                 <Typography>{t("preferences.theme")}</Typography>
                 <Switch
```
luzzif, line 44: We really should use a select here

luzzif (approved):

packages/frontend/src/components/connect-wallet/popovers/preferences/index.tsx
guerrap: Added an issue here #139.

PR #140, opened 2023-02-23
Shared state library

luzzif:
This shared state library allows sharing basic state between the host frontend and the templates. It's implemented using Redux and a series of connectors which templates can use to read state values. The library is shared between host frontend and templates through module federation. 2 main motives why this was added:

### Massively improving developer experience while developing oracles

As of now developer experience while developing KPI token templates is decent. In order to perfectly mimic real conditions, the template files are compiled (transpiled) using Carrot scripts and and Webpack and served through Webpack dev server locally, with hot module replacement active. At the same time the frontend is exported as a library to make it possible to consume locally hosted templates directly from the official Carrot frontend, giving the template developer an experience which closely resembles what he'd see in production.

The trick for KPI token templates was simply to pass an extra prop `customBaseURL` to the preview mode frontend library, so that KPI token templates code would be loaded from there and not from IPFS directly (what would happen in production), but when oracles are involved this is not enough anymore...

Oracle templates are generally picked in KPI token templates, which are not under direct control of neither the developer nor the host frontend once he's developing oracle templates (the templates seen in the app will be the actual ones present on the target chain).

This makes it necessary to have shared state between all templates at all nesting federation levels in order to pick important preferences independently from each other. In the specific case these preferences are `kpiTokenTemplateBaseURL` and `oracleTemplateBaseURL`. These 2 variables are used by the frontend when doing template content lookups and since they are now shared between templates (even of different kinds) it's trivial for oracle template developers to force KPI token templates developed by others to look for templates using a specific URL. This way oracle template developers will also have nice developer experience since they'll be able to see how their oracle components interact and integrate with different KPI token templates.

Hope it's more or less clear.


Diogomartf:
Why use Redux for state management?  

Couldn't we share state by creating a shared react context?  Seems that could reduce complexity than adding redux. 

_I also think react query might be able to handle state management but not sure if in this use case._

luzzif:
> Why use Redux for state management?
> 
> Couldn't we share state by creating a shared react context? Seems that could reduce complexity than adding redux.
> 
> _I also think react query might be able to handle state management but not sure if in this use case._

Having a React context would limit template implementors to use React as a FE tech, which is not what I'd want for long term. Besides, we use a context to share the Redux store on React frontends through the React connector.

Redux is nice because completely framework agnostic. So then we can build connectors for various FE technologies around it as needed.

Diogomartf:
> This makes it necessary to have shared state between all templates at all nesting federation levels in order to pick important preferences independently from each other

what do you mean by independently from each other?  

Isn't the idea to share the state so it's not independent from each other?

luzzif:
> > This makes it necessary to have shared state between all templates at all nesting federation levels in order to pick important preferences independently from each other
> 
> what do you mean by independently from each other?
> 
> Isn't the idea to share the state so it's not independent from each other?

Independently from each other as templates, without knowing each of the others existance, can pick up shared state. They don't have to pass state between one another directly. In that sense they are independent. There's no strict link between them, and there's no strict link to the shared state. Each part lives regardless of the other.

Diogomartf:
isn't `shared state` a little vague?   Maybe we could be more descriptive on the data being shared.


```
 {
    theme: "system",
    preferDecentralization: false,
    devMode: false,
    kpiTokenTemplateBaseURL: undefined,
    oracleTemplateBaseURL: undefined,
}
```

Some suggestions:
`shared settings` / `shared preferences`

packages/frontend/src/components/connect-wallet/popovers/preferences/index.tsx
```diff
@@ -1,7 +1,12 @@
 import React, { useCallback } from "react";
```
Diogomartf, line 1: not related to this PR, but this file doesn't make sense inside of `connect-wallet` folder, as it's not related to the connected wallet.

guerrap (approved):

packages/frontend/src/bootstrap.tsx
```diff
@@ -42,7 +42,8 @@ interface RootProps {
     providers?: ChainProviderFn[];
     connectors?: () => Connector[];
     ipfsGatewayURL?: string;
-    customBaseURL?: string;
+    kpiTokenTemplateBaseURL?: string;
+    oracleTemplateBaseURL?: string;
```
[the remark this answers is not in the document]
Diogomartf: It would be more specific, and require an URL to be passed. But it's okay with a string.

packages/frontend/src/pages/app/index.tsx
```diff
@@ -29,19 +34,28 @@ const DEFAULT_LOCATION = {
...
-    const { setPreferDecentralization } = usePreferencesSetters();
+    const setPreferDecentralization = useSetPreferDecentralization();
+    const setDevMode = useSetDevMode();
+    const setKPITokenTemplateBaseURL = useSetKPITokenTemplateBaseURL();
+    const setOracleTemplateBaseURL = useSetOracleTemplateBaseURL();
 
-    useEffect(() => {
-        if (__PREVIEW_MODE__) setPreferDecentralization(true);
-    }, [setPreferDecentralization]);
+    if (__PREVIEW_MODE__) setPreferDecentralization(true);
```
Diogomartf, line 55: if `__PREVIEW_MODE__` is a boolean we can do it like bellow.

```suggestion
   setPreferDecentralization(__PREVIEW_MODE__);
```

Diogomartf (approved):

packages/frontend/src/pages/app/index.tsx
luzzif: `__PREVIEW_MODE__` is a global defined by both Webpack and Rollup (with different values) at build time. Since the substitution is done before the optimization step it's more efficient to do it like this because when `__PREVIEW_MODE__` is false `if (__PREVIEW_MODE__) setPreferDecentralization(true);` gets rewritten to `if (false) setPreferDecentralization(true);`, which results in the whole instruction being excluded from the final bundle.

PR #147, opened 2023-02-28
Implement date picker and improve base input adding icon and action

luzzif:


Diogomartf:
looking good.

Some issues:
- on mobile it opens keyboard but you can't edit. 
- it also shows cursor which signals that the user is able to edit
image

- dark mode is missing styles

image

Diogomartf:
will we keep both Date and Datetime input?
image

luzzif:
I should have fixed all the issues in the new PR https://github.com/carrot-kpi/v1-monorepo/pull/147. And also implemented the date time picker.

PR #151, opened 2023-03-01
Account popover

luzzif:
First version of account popover and a couple other changes


packages/frontend/src/components/connect-wallet/popovers/account/index.tsx
```diff
@@ -1,37 +1,118 @@
...
+        });
         const { connectors, connect } = useConnect();
         const readonlyConnector = useMemo(() => {
             return connectors.find(
                 (connector) => connector instanceof ReadonlyConnector
             );
         }, [connectors]);
+        const blockExplorerHref = useMemo(() => {
+            if (!address || !chain || !chain.blockExplorers) return "";
+            return `${chain.blockExplorers.default.url}/address/${address}`;
```
Diogomartf, line 49: it would be great if a function return the correct url with no need for interpolation. Just pass it the address and current chain.

Could be something like this:

```suggestion
            return `blockEplorerUrl(chain, address);
```

packages/react/src/hooks/useTemplateModule.ts
```diff
@@ -25,16 +26,17 @@ export const useTemplateModule = (
...
     );
+    const ipfsGatewayURL = useIPFSGatewayURL();
 
     const { baseUrl, entry } = useMemo(() => {
         if (!template) return {};
         return {
             baseUrl:
                 customBaseURL ||
-                `${IPFSService.gateway}/ipfs/${template.specification.cid}`,
+                `${ipfsGatewayURL}/ipfs/${template.specification.cid}`,
```
Diogomartf, line 36: same, a function that given the `ipfsGatewayURL` and  `CID` return the url would be great

Diogomartf:
Just want to share my thoughts on PRs.

This PR handles more than account popover. It also handles the ipfs Gateway. 

In my opinion, It's best to do one PR focused on one thing. So in this case it would be best if was 2 different PRs.

Otherwise it makes it harder to read and understand.

Let me know your thoughts as well.

Diogomartf (approved):
Looks good, just left two improvements suggestion.

packages/frontend/src/components/connect-wallet/popovers/account/index.tsx
luzzif: I don't think Wagmi supoprts it though, no?

luzzif:
> Just want to share my thoughts on PRs.
> 
> This PR handles more than account popover. It also handles the ipfs Gateway.
> 
>

luzzif:
> Just want to share my thoughts on PRs.
> 
> This PR handles more than account popover. It also handles the ipfs Gateway.
> 
> In my opinion, It's best to do one PR focused on one thing. So in this case it would be best if it was 2 different PRs.
> 
> Otherwise it makes it harder to read and understand.
> 
> Let me know your thoughts as well.

I absolutely agree, I'll try to make PRs more focused in the future

guerrap (approved):
:rocket:

PR #152, opened 2023-03-03
Support for stateless oracle creation form

luzzif:


packages/frontend/src/pages/create-with-template-id/index.tsx
```diff
@@ -89,36 +86,15 @@ export const CreateWithTemplateId = ({
...
-        },
-        []
-    );
+    const handleCreate = useCallback(() => {
+        // a token has just been created, invalidate latest tokens query
+        queryClient.invalidateQueries(
+            // FIXME: kinda sus and ugly. The example in the docs say we
+            // can use the query prefix direcy, but it's apparently not
+            // true here
+            LATEST_KPI_TOKEN_QUERY_KEY_PREFIX as unknown as readonly unknown[]
```
Diogomartf, line 95: I don't know if this solves the problem, but we can add more info to the query key than just a string.

`useQuery({ queryKey: ['todos', { status, page }], ... })`

this might invalidate in cases that some things change.

Diogomartf (approved):

packages/frontend/src/pages/create-with-template-id/index.tsx
luzzif: No I mean in this case I'd like to invalidate all queries that have a prefix of `LATEST_KPI_TOKEN_QUERY_KEY_PREFIX` so I think the behavior is fine.

I'm not sure why though, but the library typing seems a bit wrong, since by following the docs examples this should work out of the box...

PR #156, opened 2023-03-07
Implement recent transactions list and logic to allow templates to add to it

luzzif:


packages/frontend/src/state/connector/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const useStore = createStoreHook(HostStateContext);
+export const useDispatch = createDispatchHook(HostStateContext);
+export const useSelector = createSelectorHook(HostStateContext);
+
+interface HostStateProviderProps {
+    children: ReactNode;
+}
+
+export const HostStateProvider = ({ children }: HostStateProviderProps) => {
```
Diogomartf, line 26: why the name `HostStateProvider`?

Could find out it was a `redux` provider.

packages/frontend/src/hooks/useTransactionSummary.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const [loading, setLoading] = useState(false);
+    const [summary, setSummary] = useState("");
+
+    useEffect(() => {
+        let cancelled = false;
+        const getSummary = async () => {
+            if (!cancelled) setLoading(true);
+            try {
+                const summary = await getTransactionSummary(t, provider, tx);
```
guerrap, line 21: If I'm not mistaken, `getTransactionSummary` returns a promise only when the `TxType` is `TxType.ERC20_APPROVAL`. Could it cause any issues in the other cases, since we're always expecting a promise as a result here?

packages/react/src/types/transactions.ts
```diff
@@ -0,0 +1,63 @@
...
+    transactionIndex: number;
+    blockHash: string;
+    transactionHash: string;
+    blockNumber: number;
+    status?: number;
+}
+
+export enum TxType {
+    CUSTOM,
+    KPI_TOKEN_CREATION,
```
Diogomartf, line 16: where do these variables came from?

packages/frontend/src/utils/transactions.ts
```diff
@@ -0,0 +1,144 @@
...
+
+type SummaryGetter<T extends TxType> = (
+    t: TFunction,
+    provider: providers.Provider,
+    tx: Tx<T>
+) => string | Promise<string>;
+
+// used to force implementation for all tx type variants
+// through record
+const SUMMARY_GETTER: {
```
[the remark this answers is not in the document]
luzzif: Oh no this is used to extract a i18n'd summary message from deserialized transactions

packages/react/src/types/transactions.ts
luzzif: These are determined on the base of the most used transaction types that we will most likely see in Carrot. I defined them by just knowing what the contracts do gneerally speaking.

packages/frontend/src/utils/transactions.ts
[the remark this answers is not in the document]
Diogomartf: maybe we can improve the name?

packages/frontend/src/state/connector/index.tsx
Diogomartf: If the provider is from the current repo, I think I would like more something like `CurrentStateProvider`.

But I'm okay with Host as well.

packages/frontend/src/utils/transactions.ts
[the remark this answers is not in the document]
luzzif: It sounds pretty spot on to me. The var maps tx types to summary getters, hence why `SUMMARY_GETTER`

PR #157, opened 2023-03-07
Add KPI Token campaign search with Subgraph

### LINKED ISSUE #153: Implement KPI campaign search on Subgraph
image

Diogomartf:
closes #153 

image


packages/frontend/src/pages/campaigns/top-nav/index.tsx
```diff
@@ -26,8 +33,21 @@ export const CampaignsTopNav = ({
...
-CampaignsTopNavProps) => {
+    setSearchQuery,
+}: CampaignsTopNavProps) => {
+    const [searchInputValue, setSearchInputValue] = useState<
+        string | undefined
+    >();
+
+    const handleSearchChange = useCallback(
+        (event: ChangeEvent<HTMLInputElement>): void => {
+            const inputValue = event.currentTarget.value;
```
luzzif, line 44: Unsure why `currentTarget` instead of `target`? I've always used `target` and I'm not sure about the differences.

packages/sdk/src/fetcher/subgraph/index.ts
```diff
@@ -198,15 +202,57 @@ class Fetcher implements IPartialCarrotFetcher {
             : tokens.map(({ rawAddress }) => getAddress(rawAddress));
     }
 
+    public normalizeKpiTokens = async (
+        rawTokensList: KPITokenData[] | KPITokenSearchData[],
+        chainId: ChainId,
+        ipfsGatewayURL: string
+    ) => {
+        const kpiTokens:
+            | Promise<{ [address: string]: KPIToken }>
```
luzzif, line 211: Pretty sure we could just use `{ [address: string]: KPIToken }` without the promise variant? TS should understand that.

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -8,42 +13,9 @@ import { KPITokenCard } from "../../components/ui/kpi-token-card";
 import { CampaignsTopNav } from "./top-nav";
 import { TemplatesFilter } from "./filters/templates";
 import { filterKPITokens, sortKPITokens } from "../../utils/kpi-tokens";
-import { useDebounce } from "react-use";
+import { KPIToken } from "../../../../sdk/dist";
```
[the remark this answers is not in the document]
Diogomartf: Yes. Thanks.

packages/react/src/hooks/useSearch.ts
```diff
@@ -0,0 +1,13 @@
+import { useState } from "react";
+import { useDebounce } from "react-use";
+
+export type SearchQueryProp = string | undefined;
+
+export function useSearch() {
```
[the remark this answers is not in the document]
Diogomartf: It will work if we add search to other apps. But if we can move this to frontend. I think the code is better on its own function.

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -71,15 +43,38 @@ const STATE_OPTIONS = [
...
+    // page settings
+    const [pageSize, setPageSize] = useState(12);
     const [ordering, setOrdering] = useState<SelectOption>(ORDERING_OPTIONS[0]);
     const [state, setState] = useState<SelectOption>(STATE_OPTIONS[0]);
-    const [searchQuery] = useState("");
-    const [, /* debouncedSearchQuery */ setDebouncedSearchQuery] = useState("");
+
+    // filter and sort results
+    const filteredTokens = useMemo(() => {
+        return filterKPITokens(Object.values(kpiTokens), undefined);
```
guerrap, line 65: Should we always pass undefined to filterKPITokens?

packages/react/src/hooks/useSearch.ts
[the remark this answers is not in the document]
luzzif: I'd say we could keep it and move it to the frontend for now, what do you think?

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: Don't think so. I'll remove. thanks.

packages/frontend/src/pages/campaigns/index.tsx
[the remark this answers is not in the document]
Diogomartf: Fixed.

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: Fixed.

packages/frontend/src/pages/campaigns/top-nav/index.tsx
Diogomartf: I had the types wrong before. It's working correctly. Fixed. Thanks.

packages/react/src/hooks/useSearch.ts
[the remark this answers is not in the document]
Diogomartf: I think it makes sense. I'll move it. Thanks for the feedback.

packages/sdk/src/fetcher/subgraph/index.ts
Diogomartf: I'll check.

Diogomartf: Yes. Removed. thanks.

packages/sdk/src/fetcher/subgraph/index.ts
```diff
@@ -126,15 +132,17 @@ const mapRawKPIToken = async (
...
         chainId,
         getAddress(rawKPIToken.rawAddress),
         getAddress(rawKPIToken.rawOwner),
         await mapRawTemplate(ipfsGatewayURL, rawKPIToken.template),
+
         await Promise.all(
-            rawKPIToken.oracles.map(async (rawOracle) => {
-                return mapRawOracle(chainId, ipfsGatewayURL, rawOracle);
-            })
+            rawKPIToken.oracles.map(async (rawOracle) =>
```
guerrap, line 143: I guess we can remove `async` here since we just return the promises, without `waiting` on each one, right?

guerrap:
I was testing the search a bit and it's not working properly I think. For example, I'm filtering for `Description`, and having the following campaigns I'd expect only the first one, but instead, all the campaigns get filtered out leaving no results. The description is one of the fields that we filter by.
image

packages/sdk/src/fetcher/subgraph/index.ts
Diogomartf: I'm not sure. but will check

Diogomartf:
> I was testing the search a bit and it's not working properly I think. For example, I'm filtering for `Description`, and having the following campaigns I'd expect only the first one, but instead, all the campaigns get filtered out leaving no results. The description is one of the fields that we filter by. Am I missing something? image

No, your are right, is not intended. It should filter by description.

currently we have this schema and fields include name and description. 

image

I'm trying to figure out, why that's happening. But the problem is with subgraph ([you can test it here](https://api.thegraph.com/subgraphs/name/carrot-kpi/carrot-kpi-goerli/graphql?query=query+getKPITokensBySearch%28%24query%3A+String%29+%7B%0A++kpiTokenSearch%28text%3A+%24query%29+%7B%0A++++rawAddress%3A+id%0A++++kpiToken+%7B%0A++++++rawAddress%3A+id%0A++++++owner%0A++++++descriptionCid%0A++++++description+%7B%0A++++++++title%0A++++++++description%0A++++++++tags%0A++++++%7D%0A++++++finalized%0A++++++expiration%0A++++++creationTimestamp%0A++++++oracles+%7B%0A++++++++rawAddress%3A+id%0A++++++++finalized%0A++++++%7D%0A++++++finalized%0A++++%7D%0A++%7D%0A%7D)).

We probably will need to fix/improve the subgraph.

luzzif:
The situation should be improved now, as I've changed the full text search indexing to use english as a language. Not the optimal solution, but not extremely bad either. We also need to implement the filtering while in decentralization mode, as we can't use the subgraph there. The simple solution is to, for now, fetch all KPI tokens from the blockchain and perform filtering locally (slow, but it's the tradeoff while using decentralization mode).

packages/sdk/src/fetcher/subgraph/index.ts
Diogomartf: You are right we can remove `async`. Fixed. Thanks!

guerrap (approved):

luzzif:
> > The situation should be improved now, as I've changed the full text search indexing to use english as a language. Not the optimal solution, but not extremely bad either.
> 
> It's working better. I think we can merge this for now. We can improve it more in the future, for now, I think it's okay.
> 
> > We also need to implement the filtering while in decentralization mode, as we can't use the subgraph there. The simple solution is to, for now, fetch all KPI tokens from the blockchain and perform filtering locally (slow, but it's the tradeoff while using decentralization mode
> 
> We need search on chain as well? Or just the filters?
> 
> We need to design a solution for that, I think the approach you mentioned works. I'll start to do that.

Yeah I think the proposed solution is decently good for now.

Diogomartf:
> > > The situation should be improved now, as I've changed the full text search indexing to use english as a language. Not the optimal solution, but not extremely bad either.
> > 
> > 
> > It's working better. I think we can merge this for now. We can improve it more in the future, for now, I think it's okay.
> > > We also need to implement the filtering while in decentralization mode, as we can't use the subgraph there. The simple solution is to, for now, fetch all KPI tokens from the blockchain and perform filtering locally (slow, but it's the tradeoff while using decentralization mode
> > 
> > 
> > We need search on chain as well? Or just the filters?
> > We need to design a solution for that, I think the approach you mentioned works. I'll start to do that.
> 
> Yeah I think the proposed solution is decently good for now.

I created the issues to deal with on-chain search, and the filters. I'll do other PRs.

I think we are ready to merge this PR.

PR #160, opened 2023-03-09
Implment radio input component

guerrap:


packages/ui/src/components/input/radio/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+    ref
+): ReactElement {
+    return (
+        <BaseInputWrapper
+            id={id}
+            error={error}
+            helperText={helperText}
+            className={className}
+        >
+            <div className={inputWrapperStyles({ hasLabel: !!label })}>
```
luzzif, line 77: We should pass `classNames` to internal parts to allow external styling like we do in other components.

luzzif (changes requested):
Another thing: since the `BaseInputWrapper` wraps the single radios, when setting for example an error text, it's displayed at the bottom of the single radio button. What we maybe want is instead to strip away the `BaseInputWrapper` from here and have a `RadioButtonGroup` or something that uses it instead.

luzzif (approved):

PR #162, opened 2023-03-09
Implement automatic id generation when needed and info popover

luzzif:


packages/ui/src/components/utils/popover/index.tsx
```diff
@@ -52,25 +52,34 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
...
+                placement,
+                modifiers: [
+                    {
+                        name: "offset",
+                        options: { offset },
+                    },
+                ],
+            }
+        );
+        console.log(anchor, styles, attributes, update, forceUpdate);
```
guerrap, line 68: `console.log` here.

packages/ui/src/components/utils/popover/index.tsx
```diff
@@ -52,25 +52,34 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
...
+        useLayoutEffect(() => {
+            if (update) update();
+        }, [update]);
 
         return (
             <div
-                ref={setPopper}
+                ref={(element) => {
+                    if (ref) {
+                        if (typeof ref === "function") ref(element);
```
guerrap, line 78: Why do we need to check if the `ref` is a `function` here?

luzzif: TS complains otherwise, since you can also pass functions to `ref` (that's actually what we're doing here after all with the arrow function).

packages/ui/src/components/utils/popover/index.tsx
luzzif: Addressed in 1a11900364650c00d0b8a8df8bcfa484ef6dfaaf

PR #163, opened 2023-03-09
Add info popover to radio input

guerrap:


Diogomartf:
where are we using radio buttons so I can check?

luzzif:
> where are we using radio buttons so I can check?

They're getting used in the Reality.eth oracle template, more specifically here: https://github.com/carrot-kpi/reality-eth-v3-oracle-template/blob/develop/packages/frontend/src/page/components/answer-form/index.tsx#L300

packages/ui/src/components/data-display/timer/index.tsx
```diff
@@ -53,7 +53,9 @@ export const Timer = ({
                     className={iconStyles({ className: className?.icon })}
                 />
             )}
-            {duration.format("DD[D] HH[H] mm[M] ss[S]")}
+            {duration.asSeconds() >= 0
+                ? duration.format("DD[D] HH[H] mm[M] ss[S]")
+                : "00D 00H 00M 00S"}
```
Diogomartf, line 58: I think we use this format in several places, maybe we should make this a variable and use it everywhere.

something like this:
`const ZERO_DATE_TIME_FORMAT=00D 00H 00M 00S`

luzzif:
Addressed the comment and added a `seconds` prop to the timer. By default seconds are hidden and can be shown with the prop. I've also fixed the frontend's campaign cards' timer spanning 2 lines.

PR #164, opened 2023-03-10
Remove local storage caching in dev mode

luzzif:
When testing templates this might be detrimental as it might serve old files. In production caching is handled by a service worker so we're ok.

### DIFF
--- a/packages/sdk/src/fetcher/core/index.ts
+++ b/packages/sdk/src/fetcher/core/index.ts
@@ -188,51 +188,10 @@ class Fetcher implements ICoreFetcher {
         return { ...cachedTokens, ...fetchedTokens };
     }
 
-    private async fetchContentFromIPFSWithLocalStorageCache(
-        ipfsGatewayURL: string,
-        cacheableCids: string[]
-    ) {
-        const cachedCids: { [cid: string]: string } = {};
-        const uncachedCids = [];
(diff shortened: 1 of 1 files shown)


guerrap (approved):

packages/sdk/src/fetcher/core/index.ts
```diff
@@ -188,51 +188,10 @@ class Fetcher implements ICoreFetcher {
...
-                    return {
-                        cid,
-                        content: responseOk ? await response.text() : null,
-                    };
-                })
-            );
-            for (const { cid, content } of uncachedContent) {
-                if (!content) continue;
-                cachedCids[cid] = content;
-                CACHER.set(cid, content, Number.MAX_SAFE_INTEGER);
```
guerrap, line 219: Build is failing, the `CACHER` import needs to be removed:
```
src/fetcher/core/index.ts:6:5 - error TS6133: 'CACHER' is declared but its value is never read
```

guerrap (changes requested):

packages/sdk/src/fetcher/core/index.ts
luzzif: Addressed in 13cefd5dbb020d21f97414584fcfc215b64a3a72

PR #166, opened 2023-03-13
Refactor SDK Fetcher classes

Diogomartf:
# Refactor SDK Fetcher classes
As I've been working more closer with SDK I realised that there are a lot of improvements that we can make to improve the codebase. This is an attempt to do that.

**Some problems I noticed:**

### Passing the same params all over the place
The params were used in all the functions and they were not for the specific function. They were for setup. So it makes sense to be a responsibility of the class and not the function.

We had to pass provider, ipfsGatewayURL, preferDecentralization on most functions. Like this.

```javascript
await Fetcher.fetchOracles({
      provider,
      ipfsGatewayURL,
      preferDecentralization,
      addresses: [oracleAddress],
  })
```

Now, first we initialize the class 
```javascript
const fetcher = new Fetcher({provider, ipfsGatewayURL, preferDecentralization});
```
And then we just need to pass the specific params
```javascript
await fetcher.fetchOracles({
    addresses: [oracleAddress],
})
```

I think it reads much better.  All the other params were adding noise. 

---
### Deciding the correct fetcher to use: SubgraphFetcher or OnChainFetcher, on all the functions.

In most of the functions on `FullFetcher` class we were deciding what subgraph to use, so it made sense to extract it.

```javascript
public async fetchKPITokensAmount({
        provider,
        preferDecentralization,
    }: FullFetcherFetchKPITokensAmountParams): Promise<number> {
        const useSubgraph = await this.shouldUseSubgraph({
            provider,
            preferDecentralization,
        });
        return useSubgraph
            ? SubgraphFetcher.fetchKPITokensAmount({ provider })
            : OnChainFetcher.fetchKPITokensAmount({ provider });
    }
```

Now, in the function itself it only uses a fetcher to call functions.

```javascript
    public async fetchKPITokensAmount(): Promise<number> {
        const fetcher = await this.getFetcher();
        return fetcher.fetchKPITokensAmount();
    }
```
no need to guess what the correct fetcher is, the class function `this.getFetcher()`  handles that and returns the correct fetcher.

```javascript
  private async getFetcher(): Promise<IPartialCarrotFetcher> {
        const shouldUseSubgraph = await this.shouldUseSubgraph();
        return shouldUseSubgraph ? this.subgraphFetcher : this.onChainFetcher;
    }
```
---

### A need for a place to share common functionality: BaseFetcher

The BaseFetcher class is meant to serve the base functionality for all the Fetchers classes.

```javascript

export class BaseFetcher {
    provider;
    ipfsGatewayURL;

    constructor(provider: Provider, ipfsGatewayURL: string) {
        this.provider = provider;
        this.ipfsGatewayURL = ipfsGatewayURL;
    }

    public async getChainId() {
        const { chainId } = await this.provider.getNetwork();
        enforce(chainId in ChainId, `unsupported chain with id ${chainId}`);
        return chainId;
    }

    public async getSubgraphURL(chainId?: ChainId) {
        if (!chainId) {
            chainId = await this.getChainId();
        }
        const subgraphURL = SUBGRAPH_URL[chainId];
        enforce(!!subgraphURL, `no subgraph available in chain ${chainId}`);
        return { subgraphURL };
    }

    public async getChainAddresses(chainId?: ChainId) {
        if (!chainId) {
            chainId = await this.getChainId();
        }
        const chainAddresses = CHAIN_ADDRESSES[chainId];
        enforce(!!chainAddresses, `no addresses available in chain ${chainId}`);
        return { chainAddresses };
    }
}
```

Starting by initializing the `provider` and `ipfsGatewayURL` to be accessible in all the classes.

Extracted common functionality into functions, that are used in several Fetcher classes to get chainID, subgraphURL and chain addresses.

New functions:  `getChainId`, `getSubgraphURL`, `getChainAddresses`.

To have access we just need to extend Fetcher classes to inherit from BaseFetcher like this:

```javascript
class Fetcher extends BaseFetcher implements IPartialCarrotFetcher {
```

And now we have access to those functions:

```javascript
 const chainAddresses = await this.getChainAddresses();
```

---

Also tried to improve names of functions and classes.

**Let me know what do think?** 

PS: We probably need to update the other repos using the `SDK`.  ⚠️


luzzif:
Passing the same parameters all over the place is meant to work like that. The fetcher is meant to work like a class exposing static functions, which means you can't use class variables inside, and that's why classes are made like that with no parameters in the constructors etc. 

I've only converted them to "normal" classes to have enhanced type checking through interfaces, which wasn't possible with static functions (TS doesn't support it yet).

The way I coded them mimics static member classes without the type checking limitations, I want the class to be exported once and without passed in parameters. As for the rest it could work, but I mean, I would really have preferred if you didn't waste time on this and focused instead on fixing the subgraph and make filtering work.

Diogomartf:
> Passing the same parameters all over the place is meant to work like that. The fetcher is meant to work like a class exposing static functions, which means you can't use class variables inside, and that's why classes are made like that with no parameters in the constructors etc.


When we are calling the static function, we do have everything we need to build the class and call a class function.

```js
await Fetcher.fetchOracles({
      provider,
      ipfsGatewayURL,
      preferDecentralization,
      addresses: [oracleAddress],
  })
```

We have the `provider`, `ipfsGatewayURL` and `preferDecentralization` params that are needed to create an instance of the class.

```js
const fetcher = new Fetcher({provider, ipfsGatewayURL, preferDecentralization});
```

What's the advantage of having static functions instead of class functions? 

The upside of class functions is that we have the code much more organised with less repetition and easier to read.

Diogomartf:
Potentially we can create a hook that abstracts the Fetcher class creation.

```js
export function useFetcher() {
    const preferDecentralization = usePreferDecentralization();
    const ipfsGatewayURL = useIPFSGatewayURL();
    const provider = useProvider();
    const [fetcher, setFetcher] = useState<Fetcher>(
        new Fetcher({
            provider,
            ipfsGatewayURL,
            preferDecentralization,
        })
    );

    useEffect(() => {
        setFetcher(
            new Fetcher({
                provider,
                ipfsGatewayURL,
                preferDecentralization,
            })
        );
    }, [provider, ipfsGatewayURL, preferDecentralization]);

    return { fetcher };
}
```

And then use it like this:

image

Diogomartf:
I added a `FetcherProvider` so it's always the same instance of the class. (previously we were instanciating the class when router changes).

Integrated the `FetcherProvider` with `CarrotCoreProvider` so we can access fetcher on repos with that provider. 
Also FetcherProvider is available for those who just want access the Fetcher class.

I kept the `useFetcher` to make it very easy  to use it, it was updated to use the provider context, `FetcherContext`.

I think overall we kept the performance and improved the code design. 

Did I not address any of the concerns you raised @luzzif?  Is there anything missing? I would like to see this code merged. 🙌

Diogomartf:
I think overall we kept the performance and improved the code design. 

Did I not address any of the concerns you raised @luzzif?  Is there anything missing? 

I would like to see this code merged. 🙌

PR #167, opened 2023-03-14
Gnosis chain support

luzzif:
Since Goerli is expensive right now and I can't deploy the newest version of the contracts, this PR adds Gnosis chain support.


Diogomartf (approved):
Looks good to me 👍

PR #170, opened 2023-03-15
Disabled network switching while in modal mode

luzzif:
In modal mode (i.e. when consuming a template) network switching is now disabled.


packages/frontend/src/components/connect-wallet/index.tsx
```diff
@@ -65,6 +78,8 @@ export const ConnectWallet = () => {
         setAccountPopoverOpen(false);
     }, []);
 
+    console.log(modalNetworksPopoverOpen);
```
guerrap, line 81: wild `console.log`

PR #175, opened 2023-03-15
Multichain URLs

luzzif:
This PR implements multichain URLs that result in shareable multichain links.

The behavior should be:

- When you land on Carrot and no chain is in the URL, the currently active chain in the dapp is set in the URL.
- When you land on Carrot and a chain is specified in the URL, there are 2 possibilities:
    - If the currently active chain and the URL chain match, nothing happens, the user should see the page normally.
    - If there's a mismatch, the user is asked to switch network in his wallet in order to see the page.


packages/frontend/src/components/connect-wallet/index.tsx
```diff
@@ -34,6 +46,85 @@ export const ConnectWallet = ({ mode }: ConnectWalletProps) => {
...
+            setSearchParams(searchParams);
+            return;
+        }
+
+        // if the active connector is readonly, always switch to the target chain
+        const readonlyConnectorActive =
+            activeConnector instanceof ReadonlyConnector;
+        if (readonlyConnectorActive) {
+            const targetChainId = chains.find(
+                (c) => c.name.toLowerCase() === targetChainName.toLowerCase()
```
Diogomartf, line 88: I think calling the variable chain is more clear:

```suggestion
                (chain) => chain.name.toLowerCase() === targetChainName.toLowerCase()
```

packages/frontend/src/components/connect-wallet/index.tsx
```diff
@@ -34,6 +46,85 @@ export const ConnectWallet = ({ mode }: ConnectWalletProps) => {
...
+        setChainFromSearchParams(targetChain);
+    }, [canFreelySwitchNetworks, chains, searchParams]);
+
+    // this effect does a couple things:
+    // - in case the dapp's url has no chain search param, it sets it to the currently connected chain
+    // - if a target chain is specified, but we're in the middle of a network switching event, it does nothing
+    // - if a target chain is specified, it is different to the currently active chain, and we're not in the
+    // middle of a network switching event, it shows the wrong network modal, inviting the user to switch the
+    // network in their wallet (it does so only the first time)
+    useEffect(() => {
```
Diogomartf, line 74: it would be great if we could break this big `useEffect`

Diogomartf:
Not sure if it is related to this PR, but I'm not able to switch networks when my wallet current network is unsupported.

image

Clicking in a network doesn't ask my wallet to switch networks.

packages/frontend/src/components/connect-wallet/index.tsx
luzzif: I think in the latest push I've done that, I've overall simplified the logic a bit

luzzif:
> Not sure if it is related to this PR, but I'm not able to switch networks when my wallet (Tally/taho) current network is unsupported.
> 
> image
> 
> Clicking in a network doesn't ask my wallet to switch networks.
> 
> **updates**
> 
> This works fine on Metamask.

Weid... which wallet are you using?

Diogomartf:
> > Not sure if it is related to this PR, but I'm not able to switch networks when my wallet (Tally/taho) current network is unsupported.
> 
> > 
> 
> > image
> 
> > 
> 
> > Clicking in a network doesn't ask my wallet to switch networks.
> 
> > 
> 
> > **updates**
> 
> > 
> 
> > This works fine on Metamask.
> 
> 
> 
> Weid... which wallet are you using?

Taho (ex. Tally ho)

But works fine with metamask.

luzzif:
> > > Not sure if it is related to this PR, but I'm not able to switch networks when my wallet (Tally/taho) current network is unsupported.
> > 
> > 
> > > 
> > 
> > 
> > > image
> > 
> > 
> > > 
> > 
> > 
> > > Clicking in a network doesn't ask my wallet to switch networks.
> > 
> > 
> > > 
> > 
> > 
> > > **updates**
> > 
> > 
> > > 
> > 
> > 
> > > This works fine on Metamask.
> > 
> > 
> > Weid... which wallet are you using?
> 
> Taho (ex. Tally ho)
> 
> But works fine with metamask.

I tried with Taho too and it looks like it fails because Taho isn't supporting Sepolia nor Gnosis. And it doesn't even support adding custom networks and probably doesn't support automatic network addition too. So it's more about the wallet I think.

guerrap (approved):
:rocket:

PR #178, opened 2023-03-16
Implement frame connector

luzzif:


packages/frontend/src/connectors/frame/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        return provider.getSigner();
+    }
+
+    async isAuthorized() {
+        return true;
+    }
+
+    async switchChain(chainId: number): Promise<Chain> {
+        await this.getProvider({ chainId, create: true });
+        this.onChainChanged(chainId);
```
guerrap, line 106: Should we move `this.onChainChanged(chainId);` below the if `(!chain) ...`?

guerrap (approved):

packages/frontend/src/connectors/frame/index.ts
luzzif: U rite, and I've also fixed the readonly connector in1a52d868580bfc72e64f48597a2fbec0dd25640b

PR #183, opened 2023-03-17
fix(frontend,ui): make popover update position automatically by avoiding the use of refs

luzzif:


packages/frontend/src/components/connect-wallet/index.tsx
```diff
@@ -10,36 +10,49 @@ import { ChainIcon } from "../chain-icon";
...
-    useClickAway(connectPopoverRef, () => {
-        setConnectPopoverOpen(false);
-    });
-
-    useClickAway(accountPopoverRef, () => {
-        setAccountPopoverOpen(false);
-    });
-
+    useEffect(() => {
+        const handleClick = (event: MouseEvent) => {
```
guerrap, line 34: Should we rename this `handleMouseDown` since it's reacting to `mousedown` events?

guerrap (approved):

packages/frontend/src/components/connect-wallet/index.tsx
luzzif: Addressed in 7fbbd22342ede86656644d348431278e24706b30

PR #185, opened 2023-03-20
Implement kpi campaign search on chain

Diogomartf:
Make search work when decentralize mode is on.

image

### DIFF
--- a/packages/sdk/src/fetcher/search/index.ts
+++ b/packages/sdk/src/fetcher/search/index.ts
@@ -1,0 +1,23 @@
+import { KPIToken } from "../../entities/kpi-token";
+import { KPITokensObjectProp, transformInKPITokensObject } from "../serializer";
+
+export function searchKPItokens(
+    searchQuery: string,
+    allKPITokens: KPITokensObjectProp
+) {
+    const filteredTokens = Object.values(allKPITokens).filter((token) => {
+        return tokenSpecificationIncludesQuery(token, searchQuery);
+    });
+
+    return transformInKPITokensObject(filteredTokens);
+}
+
+function tokenSpecificationIncludesQuery(token: KPIToken, searchQuery: string) {
+    const query = searchQuery.toLowerCase();
+    return (
+        token.hasOwnProperty("specification") &&
+        (token.specification.title.toLowerCase().includes(query) ||
+            token.specification.description.toLowerCase().includes(query) ||
+            Object.values(token.specification.tags).includes(query))
+    );
+}
--- a/packages/sdk/src/fetcher/search/index.ts
+++ b/packages/sdk/src/fetcher/search/index.ts
@@ -1,9 +1,9 @@
 import { KPIToken } from "../../entities/kpi-token";
-import { KPITokensObjectProp, transformInKPITokensObject } from "../serializer";
+import { KPITokensObject, transformInKPITokensObject } from "../serializer";
 
 export function searchKPItokens(
     searchQuery: string,
-    allKPITokens: KPITokensObjectProp
+    allKPITokens: KPITokensObject
 ) {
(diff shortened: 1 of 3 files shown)


packages/sdk/src/fetcher/serializer/index.ts
```diff
@@ -0,0 +1,15 @@
+import { KPIToken } from "../../entities/kpi-token";
+
+export type KPITokensObjectProp = { [address: string]: KPIToken };
```
guerrap, line 3: I think we can rename this to `KPITokensObjectProp` since it's not used as a prop for a component,
Also, we could use the `Record` utility type here, something like:

```
export type KPITokensObject = Record<string, KPIToken>;
```

Wdyt?

Diogomartf: I think it's a good suggestion, will update. Thanks!

packages/sdk/src/fetcher/search/index.ts
```diff
@@ -0,0 +1,23 @@
...
+    return transformInKPITokensObject(filteredTokens);
+}
+
+function tokenSpecificationIncludesQuery(token: KPIToken, searchQuery: string) {
+    const query = searchQuery.toLowerCase();
+    return (
+        token.hasOwnProperty("specification") &&
+        (token.specification.title.toLowerCase().includes(query) ||
+            token.specification.description.toLowerCase().includes(query) ||
+            Object.values(token.specification.tags).includes(query))
```
guerrap, line 21: I think we should also make the tags lowercase `token.specification.tags.toString().toLowerCase()`.
Also, is `Object.values()` actually needed?

Diogomartf: Done

guerrap (approved):

packages/sdk/src/fetcher/search/index.ts
```diff
@@ -0,0 +1,40 @@
...
+) {
+    const filteredTokens = Object.values(allKPITokens).filter((token) =>
+        tokenSpecificationIncludesQuery(token, searchQuery)
+    );
+
+    return transformInKPITokensObject(filteredTokens);
+}
+
+function tokenSpecificationIncludesQuery(token: KPIToken, searchQuery: string) {
+    if (!token.hasOwnProperty("specification")) return;
```
luzzif, line 16: Why is this a check?

packages/sdk/src/fetcher/serializer/index.ts
```diff
@@ -0,0 +1,15 @@
+import { KPIToken } from "../../entities/kpi-token";
+
+export type KPITokensObject = Record<string, KPIToken>;
```
luzzif, line 3: A bit unneeded to me to have this instead of directly writing `Record<string, KPIToken>`, but not anything that is blocking at all.

packages/sdk/src/fetcher/search/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const query = searchQuery.toLowerCase();
+    const specification = token.specification;
+
+    return (
+        token.hasOwnProperty("specification") &&
```
luzzif, line 22: This is a duplicate check.

packages/sdk/src/fetcher/search/index.ts
Diogomartf: I think we can remove the check. Thanks for pointing out.

packages/sdk/src/fetcher/search/index.ts
Diogomartf: Will fix. thanks

Diogomartf:
I will close this PR and open a new one pointing the merge on `develop` branch.

PR #189, opened 2023-03-21
feat: add campaigns sorting by date (newest and oldest)

Diogomartf:
3c8f5fec918821f691659b610bdc643d

### DIFF
--- a/packages/frontend/src/pages/campaigns/index.tsx
+++ b/packages/frontend/src/pages/campaigns/index.tsx
@@ -14,11 +14,11 @@ import { useSearch } from "../../hooks/useSearch";
 
 const ORDERING_OPTIONS = [
     {
-        label: "Latest",
+        label: "Newest",
         value: 1,
     },
     {
-        label: "Newest",
+        label: "Oldest",
         value: 2,
     },
 ];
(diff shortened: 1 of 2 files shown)


packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -56,14 +56,14 @@ export const Campaigns = () => {
...
+    // filter results
     const filteredTokens = useMemo(() => {
         return filterKPITokens(Object.values(kpiTokens));
     }, [kpiTokens]);
-
+    // sort results
     const sortedFilteredTokens = useMemo(() => {
-        return sortKPITokens(filteredTokens);
-    }, [filteredTokens]);
+        return sortKPITokens(filteredTokens, ordering.label);
```
guerrap, line 65: Maybe it's clearer to pass the `value` to the `sortKPITokens`; this way we avoid the `ordering?.toLowerCase() === "oldest"`.

One more thing, I'd use the values `-1` as `descending` and `1` as `ascending` for the sorting.

luzzif: Agree on both points. We can also work with strings directly if we make the values `newest` and `oldest` respectively instead of `1` and `2`, and then the `sortKPITokens` can take them in. A bit like it was before. I think it's way more elegant.

luzzif (changes requested):

Diogomartf:
I will close this PR, and added a new one pointing to `develop`.

https://github.com/carrot-kpi/v1-monorepo/pull/197

PR #194, opened 2023-03-22
Implement KPI campaigns search on-chain with decentralize mode

Diogomartf:
When enabling decentralize mode, we connect directly on chain contract to get the data instead of subgraph. 

This PR adds search support to the on chain fetcher.

image

This is related to this PR: #185


luzzif (approved):

MilanVojnovic95:
Tested and looks ok to me 👍

PR #197, opened 2023-03-22
feat: add campaigns sorting by date

### LINKED ISSUE #187: Implement order filters
image

Diogomartf:
closes #187 

3c8f5fec918821f691659b610bdc643d

Related to this PR: #189

### DIFF
--- a/packages/frontend/src/pages/campaigns/index.tsx
+++ b/packages/frontend/src/pages/campaigns/index.tsx
@@ -12,14 +12,19 @@ import { Empty } from "../../components/ui/empty";
 import { KPIToken } from "@carrot-kpi/sdk";
 import { useSearch } from "../../hooks/useSearch";
 
+export enum Order {
+    newest,
+    oldest,
+}
+
 const ORDERING_OPTIONS = [
     {
         label: "Newest",
-        value: 1,
+        value: Order.newest,
     },
     {
         label: "Oldest",
-        value: 2,
+        value: Order.oldest,
     },
 ];
 
--- a/packages/frontend/src/pages/campaigns/index.tsx
+++ b/packages/frontend/src/pages/campaigns/index.tsx
@@ -38,11 +38,6 @@ const STATE_OPTIONS = [
     },
 ];
 
-export interface OrderOption {
-    label: string;
-    value: number;
-}
-
 export const Campaigns = () => {
     useResetPageScroll();
     const { t } = useTranslation();
@@ -58,7 +53,7 @@ export const Campaigns = () => {
 
     // page settings
     const [pageSize, setPageSize] = useState(12);
-    const [ordering, setOrdering] = useState<OrderOption>(ORDERING_OPTIONS[0]);
+    const [ordering, setOrdering] = useState<SelectOption>(ORDERING_OPTIONS[0]);
     const [state, setState] = useState<SelectOption>(STATE_OPTIONS[0]);
 
     // filter results
@@ -67,7 +62,7 @@ export const Campaigns = () => {
     }, [kpiTokens]);
     // sort results
     const sortedFilteredTokens = useMemo(() => {
-        return sortKPITokens(filteredTokens, ordering.value);
+        return sortKPITokens(filteredTokens, Number(ordering.value));
     }, [filteredTokens, ordering]);
 
     useEffect(() => {
--- a/packages/frontend/src/pages/campaigns/index.tsx
+++ b/packages/frontend/src/pages/campaigns/index.tsx
@@ -11,6 +11,7 @@ import { filterKPITokens, sortKPITokens } from "../../utils/kpi-tokens";
 import { Empty } from "../../components/ui/empty";
 import { KPIToken } from "@carrot-kpi/sdk";
 import { useSearch } from "../../hooks/useSearch";
+import { t } from "i18next";
 
 export enum Order {
     newest,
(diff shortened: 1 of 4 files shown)


packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -14,11 +14,11 @@ import { useSearch } from "../../hooks/useSearch";
 
 const ORDERING_OPTIONS = [
     {
-        label: "Latest",
+        label: "Newest",
         value: 1,
```
luzzif, line 18: I'd still highly prefer to see these as strings as it's much clearer in the code.

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -38,6 +38,11 @@ const STATE_OPTIONS = [
     },
 ];
 
+export interface OrderOption {
```
luzzif, line 41: Not sure why we need this? Can't we use the `SelectOption` exported by the `Select` component in the UI lib it should be the exact same data structure and the name difference doesn't increase readability that much.

luzzif (changes requested):

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: it was complaining because the value can be a string on SelectOption. I guess I can cast it somewhere and use the select option,

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: what do you mean by seeing this as strings, aren't they strings?

packages/frontend/src/pages/campaigns/index.tsx
luzzif: Ohhh yeah, we can use the casting. Admittedly, we can improve typings on the `Select`

packages/frontend/src/pages/campaigns/index.tsx
luzzif: No I mean the values. Values could be `newest-first` and `latest-first` or something like that and then in the sorting function you can type the ordering to be `"newest-first" | "oldest-first"` so that it's overall a bit clearer.

This makes us avoid using numbers as ordering options and makes the solution a bit more expressive.

Also, we really should support i18n in the select options through the `t` function.

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: fixed: a26cf15c061aa03f79ceb3d7f5e489fdc74ef5c7

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: so instead the value being 1 it should be something like "newest-first"?

Diogomartf: I'll add i18n support, tks.

Diogomartf: Typescript has support for enums. We can do something like this:

image

luzzif: Yeah but I find that to be a more convoluted way of just having the strings in a way

Diogomartf: Let's use this for now. Updated here: bbdc10b8a89e819d5c2c99714389ec3bb4fa6352

Diogomartf: Added translations as well: 7b03fc9ade46d439806daf392105c4e4792e4e10

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -11,15 +11,21 @@ import { filterKPITokens, sortKPITokens } from "../../utils/kpi-tokens";
 import { Empty } from "../../components/ui/empty";
 import { KPIToken } from "@carrot-kpi/sdk";
 import { useSearch } from "../../hooks/useSearch";
+import { t } from "i18next";
+
+export enum CampaignOrder {
+    newest,
```
luzzif, line 17: Enum entries are usually all uppercase

luzzif (changes requested):

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: True, thanks

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -56,14 +62,14 @@ export const Campaigns = () => {
...
+    // filter results
     const filteredTokens = useMemo(() => {
         return filterKPITokens(Object.values(kpiTokens));
     }, [kpiTokens]);
-
+    // sort results
     const sortedFilteredTokens = useMemo(() => {
-        return sortKPITokens(filteredTokens);
-    }, [filteredTokens]);
+        return sortKPITokens(filteredTokens, Number(ordering.value));
```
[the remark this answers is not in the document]
Diogomartf: Argument of type 'ValueType' is not assignable to parameter of type 'CampaignOrder'.

I think we still need to cast it.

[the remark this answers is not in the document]
Diogomartf: I let the  cast function`number()`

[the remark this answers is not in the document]
luzzif: No wait, do `ordering.value as unknown as CampaignOrder`. That works. Not super pretty but better than casting to number imho

luzzif (approved):

PR #199, opened 2023-03-22
Use flexbox instead of grid

guerrap:
Fixes https://github.com/carrot-kpi/erc20-kpi-token-template/issues/74.


Diogomartf (approved):

guerrap:
> this is much better!
> 
> I would say that aligning left (or removing center alignment) looks even better (maybe add some padding x).
> 
> image

I've made a small change, I think we're good with this for now :slightly_smiling_face:

PR #202, opened 2023-03-23
fix(frontend): only show frame connector when frame is actually available

luzzif:
Closes https://github.com/carrot-kpi/erc20-kpi-token-template/issues/67

### DIFF
--- a/packages/frontend/src/connectors/frame/index.ts
+++ b/packages/frontend/src/connectors/frame/index.ts
@@ -25,7 +25,6 @@ export class FrameConnector extends Connector<JsonRpcProvider, unknown> {
             }),
         })
             .then((response) => {
-                console.log(response.ok);
                 this.ready = response.ok;
             })
             .catch((error) => {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/connectors/frame/index.ts
```diff
@@ -7,12 +7,34 @@ export const FRAME_CONNECTOR_ID = "frame";
...
+            method: "POST",
+            body: JSON.stringify({
+                jsonrpc: "2.0",
+                method: "eth_blockNumber",
+                params: [],
+                id: 1,
+            }),
+        })
+            .then((response) => {
+                console.log(response.ok);
```
guerrap, line 28: Should we keep this log?

guerrap (dismissed):

packages/frontend/src/connectors/frame/index.ts
```diff
@@ -7,12 +7,34 @@ export const FRAME_CONNECTOR_ID = "frame";
...
-    readonly ready = true;
+    ready: boolean;
 
     private provider?: JsonRpcProvider;
 
     constructor(config: { chains: Chain[]; options: object }) {
         super(config);
+        this.ready = false;
+
+        fetch("http://127.0.0.1:1248", {
```
Diogomartf, line 18: is this the right url?

packages/frontend/src/connectors/frame/index.ts
luzzif: Addressed in dd1a993b3ce0134ddf8e38983ade22c7acb6e814

PR #203, opened 2023-03-23
feat(sdk): make library tree shakeable

luzzif:
The SDK library got restructured. Now it exports 2 main folders:

- `./dist` contains a bundled version of the library that WON'T be tree shaken
- `./lib` exports separated modules as present in the library itself. By importing only what you need, this enables tree shaking to be performed by bundlers.

The new structure is already applied to internal usage of the SDK.

Gotta say I'm not a MASSIVE fan of this solution so if you have advices please give them here.


packages/sdk/src/fetcher/on-chain.ts
```diff
@@ -10,48 +11,52 @@ import {
...
 // platform related interfaces
 const KPI_TOKEN_INTERFACE = new Interface(KPI_TOKEN_ABI);
 const ORACLE_INTERFACE = new Interface(ORACLE_ABI);
 const KPI_TOKENS_MANAGER_INTERFACE = new Interface(KPI_TOKENS_MANAGER_ABI);
 
 // platform related functions
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
 const KPI_TOKEN_TEMPLATE_FUNCTION =
-    KPI_TOKEN_INTERFACE.getFunction("template()");
+    KPI_TOKEN_INTERFACE.getFunction("template()")!;
```
Diogomartf, line 39: why add a `!`?

packages/sdk/src/fetcher/index.ts
Diogomartf, line 1: why do we need to put `.js` extension?

PR #206, opened 2023-03-24
feat: add checkboxes filters

### LINKED ISSUE #205: implement Template filters
image

Diogomartf:
closes #205 

Added `CheckboxesFilter` component.

```javascript
 <CheckboxesFilter
      title="Oracles"
      groupId="campaigns-oracles"
      loading={loading}
      items={templates}
      selected={selectedOracles}
      setSelected={setSelectedOracles}
  />
```

It manages
- the layout (title, styles, loadings) 
- handles checkboxes change
- handles showing the correct checkboxes

`selected` is a `Set` of selected `addresses` 

image

which we use to filter

```javascript
 selectedTemplates.has(token.template.address) 
```

**Preview**
This screenshot is slate. Now the default is show all data and by checking we filter the data.

40542e88dc6c9630e2257614ac121ea7


packages/frontend/src/pages/campaigns/filters/utils/index.ts
```diff
@@ -0,0 +1,38 @@
...
+export const CARROT_DEFAULT_TEMPLATE_ID = 1;
+export const CARROT_DEFAULT_ORACLE_TEMPLATE = 1;
+
+type CheckboxProp = {
+    id: number;
+    name: string;
+    checked: boolean;
+};
+
+export type CheckboxesObj = Record<number, CheckboxProp>;
```
guerrap, line 12: Should we rename this to just `Checkboxes`?

packages/frontend/src/pages/campaigns/filters/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+                selectedOracles={selectedOracles}
+            />
+        </div>
+        <Button
+            className={{
+                root: "w-full p-1 mt-12 md:hidden",
+            }}
+            onClick={toggleFilters}
+        >
+            Close filters
```
guerrap, line 62: We should add this to `i18n`.

packages/frontend/src/pages/campaigns/filters/templates/index.tsx
```diff
@@ -1,34 +1,30 @@
...
-                                id={template.id.toString()}
-                                label={template.specification.name}
-                            />
-                        );
-                    })
-                )}
-            </div>
-        </div>
+        <CheckboxesFilter
+            title="Templates"
```
guerrap, line 21: `i18n`

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+    setSelected,
+    defaultChecked,
+    groupId,
+}: CheckboxesFilterProps) => {
+    const [checkboxes, setCheckboxes] = useState<CheckboxesObj>({});
+
+    const initializeCheckboxes = useCallback(() => {
+        const hasItems = items.length !== 0;
+        const hasCheckboxes = Object.keys(checkboxes).length !== 0;
+
```
guerrap, line 36: Maybe we can rewrite this as:

```ts
if(items.length === 0) return;
if(Object.keys(checkboxes).length !== 0) return;
```

I think it's a bit more clear.

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+                defaultChecked
+            );
+            setCheckboxes(defaultCheckboxes);
+            setSelected(getDefaultSelected(defaultCheckboxes));
+        }
+    }, [checkboxes, defaultChecked, items, setSelected]);
+
+    const handleCheckedChange = (e: ChangeEvent<HTMLInputElement>) => {
+        const element = e.target;
+        const elementId = Number(e.target.getAttribute("data-id"));
```
guerrap, line 49: Here we can use `element`, so `element.getAttribute(...)`

luzzif, line 49: A more robust way to handle this is:

```
const id = (event.target as HTMLLIElement).dataset.id;
if (id !== undefined)
    const parsedId = parseInt(id);
    if (!isNaN(parsedId)) {
        ...
    }
}
```

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+        newCheckboxes[elementId] = {
+            ...newCheckboxes[elementId],
+            checked: element.checked,
+        };
+        setCheckboxes(newCheckboxes);
+
+        const newSelected = selected;
+
+        if (element.checked === true) newSelected.add(elementId);
+        if (element.checked === false && selected.has(elementId))
```
luzzif, line 61: This can just be more efficiently rewritten as:

```
if (element.checked) newSelected.add(elementId);
else if (selected.has(elementId)) newSelected.delete(elementId);
```

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+            newSelected.delete(elementId);
+
+        setSelected(newSelected);
+    };
+
+    useEffect(() => {
+        initializeCheckboxes();
```
luzzif, line 68: If a function is used only one time and is not used to split up logic, I generally think it doesn't make much sense to have it. This is one of those cases. Also, we avoid a `useCallback` which React has to manage.

packages/frontend/src/pages/campaigns/filters/oracles/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const OraclesFilter = ({
+    selectedOracles,
+    setSelectedOracles,
+}: {
+    selectedOracles: Set<number>;
+    setSelectedOracles: (newSelectedOracles: Set<number>) => void;
+}) => {
+    const { loading, templates } = useOracleTemplates();
+
+    const defaultChecked = (item: Template) =>
```
luzzif, line 16: Not sure I understand this and why it is a function.

packages/frontend/src/pages/campaigns/filters/templates/index.tsx
```diff
@@ -1,34 +1,30 @@
...
+export const TemplatesFilter = ({
+    selectedTemplates,
+    setSelectedTemplates,
+}: {
+    selectedTemplates: Set<number>;
+    setSelectedTemplates: (newSelectedTemplates: Set<number>) => void;
+}) => {
     const { loading, templates } = useKPITokenTemplates();
 
+    const defaultChecked = (item: Template) =>
```
luzzif, line 16: Same as above

packages/frontend/src/pages/campaigns/filters/utils/index.ts
```diff
@@ -0,0 +1,38 @@
+import { Template } from "@carrot-kpi/sdk";
+
+export const CARROT_DEFAULT_TEMPLATE_ID = 1;
```
luzzif, line 3: I'm not sure I fully get what these are for. It should be the template id for the default selected checkboxes right? If that's the case, why is this a thing? I'd imagine the default state of the checkboxes on page render is for all of them to be selected (as we would show ALL the campaigns using any of the templates in that case).

packages/frontend/src/pages/campaigns/filters/utils/index.ts
luzzif: I don't necessarily think it makes tooooo much sense to have this type when it's simply a rename, but I can see it having some value in refactoring scenarios (where you can only change the value of the type here and it changes everywhere).

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -77,25 +64,72 @@ export const Campaigns = () => {
...
+
+        return filterStateKPITokens(
+            tokens,
             campaignState.value as unknown as CampaignState
         );
     }, [kpiTokens, campaignState]);
 
+    const hasOracleTemplate = useMemo(
+        () => (token: KPIToken) =>
+            oraclesTemplateIds(token).some((oracleId) => {
```
luzzif, line 111: Looks like the `oraclesTemplateIds` function below is only used here, so I think we can just write this as:

```
token.oracles.map((oracle) => oracle.template.id).some((oracleId) => {
    return selectedOracles.has(Number(oracleId));
})
```

luzzif: Also, I'm not sure why the `Number(...)` casts everywhere on the template id. It should already be of type number.

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -77,25 +64,72 @@ export const Campaigns = () => {
...
+        [selectedTemplates]
+    );
+
+    const oraclesTemplateIds = (token: KPIToken) =>
+        token.oracles.map((oracle) => oracle.template.id);
+
+    const filteredTokens = useMemo(
+        () =>
+            filteredKPITokensByState.filter(
+                (token) => hasTemplate(token) || hasOracleTemplate(token)
```
luzzif, line 127: Same as above. The `hasTemplate` and `hasOracleTemplate` functions are only used here, so it doesn't make sense to have them. Let's move all of their logic in this `useMemo`. We also make React's job easier as we slim down the list of dependencies on the `useMemo` hook and overall reduce the `useMemo` usage.

luzzif (changes requested):

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
Diogomartf: Yes. Thaaaanks

packages/frontend/src/pages/campaigns/filters/index.tsx
Diogomartf: I'll add it

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
Diogomartf: Ok. I'll change that.

packages/frontend/src/pages/campaigns/filters/oracles/index.tsx
Diogomartf: this function is to tell the `CheckboxesFilter` component what are the checkboxes that we want to be checked by default.

packages/frontend/src/pages/campaigns/filters/utils/index.ts
Diogomartf: So basically we should show all by default and only filter when checked.

That makes sense. 

Yup, in that case this isn't needed. 

Thanks for pointing this out.

packages/frontend/src/pages/campaigns/filters/oracles/index.tsx
Diogomartf: We do not need this function because:

https://github.com/carrot-kpi/v1-monorepo/pull/206/files#r1152053453

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: I'll take a look on this. But compiler issues.

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
Diogomartf: fixed: https://github.com/carrot-kpi/v1-monorepo/pull/206/commits/c6a5a7f110066cc7fc709404d55e4b1d15027682

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
Diogomartf: fixed 0b0e3127244ed8bb7ede963d6eb0868f403cc73a

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
Diogomartf: Much better. Thanks!

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: `Number` is not needed in this case.

packages/frontend/src/pages/campaigns/filters/utils/index.ts
luzzif: `CheckboxPropsMap` sounds decent, no?

packages/frontend/src/pages/campaigns/filters/utils/index.ts
luzzif: LESGOOOOOOOOOOOo

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const initializeCheckboxes = useCallback(() => {
+        if (items.length === 0) return;
+        if (Object.keys(checkboxes).length !== 0) return;
+
+        const defaultCheckboxes = getDefaultCheckboxes(items, defaultChecked);
+        setCheckboxes(defaultCheckboxes);
+        setSelected(getDefaultSelected(defaultCheckboxes));
+    }, [checkboxes, defaultChecked, items, setSelected]);
+
+    const getHtmlDataId = (element: HTMLInputElement) => {
```
luzzif, line 42: This function not being in a `useCallback` is a bit of an issue on rerenders. See comment below.

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+        if (id !== undefined) {
+            const parsedId = parseInt(id);
+            if (!isNaN(parsedId)) {
+                return parsedId;
+            }
+        }
+        return 0;
+    };
+
+    const handleCheckedChange = (e: ChangeEvent<HTMLInputElement>) => {
```
luzzif, line 54: This and the previous functions not using `useCallback` cause a rerender on the checkbox each single time this component is called. I also think we can directly write the `getHtmlDataId` logic here to avoid an extra dependency in this `useCallback`able function.

luzzif (changes requested):

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
Diogomartf: I remove it.

packages/frontend/src/pages/campaigns/filters/utils/index.ts
Diogomartf: fixed: df180fc782eaa6d8a035701179412777fc890cba

packages/frontend/src/pages/campaigns/filters/utils/index.ts
Diogomartf: We ended up not needing it.

fixed: df180fc782eaa6d8a035701179412777fc890cba

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: fixed: oraclesTemplateIds

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: fixed: df180fc782eaa6d8a035701179412777fc890cba

packages/frontend/src/pages/campaigns/filters/index.tsx
Diogomartf: fixed: 1b624c21dc7e310b7084ee6a821db29912b08d09

packages/frontend/src/pages/campaigns/filters/oracles/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+        item.id === CARROT_DEFAULT_ORACLE_TEMPLATE;
+
+    return (
+        <CheckboxesFilter
+            title="Oracles"
```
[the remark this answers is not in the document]
Diogomartf: fixed: 1b624c21dc7e310b7084ee6a821db29912b08d09

packages/frontend/src/pages/campaigns/filters/templates/index.tsx
Diogomartf: fixed: 1b624c21dc7e310b7084ee6a821db29912b08d09

packages/frontend/src/pages/campaigns/filters/checkboxes-filter/index.tsx
Diogomartf: I added the `getHtmlDataId` logic directly there, also deleted the utils since became almost empty.

Diogomartf: I'm not getting the `useCallback`. It's not suppose the handleCheckbox to run every time a checkbox changes state?

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -88,25 +38,60 @@ export const Campaigns = () => {
...
             searchedResolvedKPITokens,
             campaignState.value as unknown as CampaignState
         );
     }, [searchedResolvedKPITokens, campaignState]);
 
+    const filteredTokens = useMemo(() => {
+        const noFiltersSelected =
+            selectedTemplates.size === 0 && selectedOracles.size === 0;
+
+        if (noFiltersSelected) return filteredKPITokensByState;
```
guerrap, line 84: I think we can remove `noFiltersSelected` and put the condition directly in the `if`.

PR #207, opened 2023-03-25
Error boundaries

guerrap:
Wrap the `TemplateComponent` with `ErrorBoundary`; this way we avoid the error propagation when a federated component fails to load.


packages/frontend/src/pages/create-with-template-id/index.tsx
```diff
@@ -138,6 +138,20 @@ export const CreateWithTemplateId = ({
                                 <Loader />
                             </div>
                         }
+                        error={
+                            <div className="bg-green py-10 flex justify-center">
+                                <ErrorFeedback
+                                    messages={{
+                                        title: i18n.t(
```
luzzif, line 145: Why use `i18n.t`? Can't we directly use `t`?

packages/frontend/src/pages/page/index.tsx
```diff
@@ -67,6 +67,20 @@ export const Page = ({ closing, onOutAnimationEnd }: PageProps) => {
                                 <Loader />
                             </div>
                         }
+                        error={
+                            <div className="bg-orange py-10 flex justify-center">
+                                <ErrorFeedback
+                                    messages={{
```
luzzif, line 73: The above also applies here

packages/react/package.json
```diff
@@ -30,6 +30,7 @@
         "@tanstack/react-query-persist-client": "^4.24.9",
         "@uniswap/token-lists": "1.0.0-beta.31",
         "class-variance-authority": "^0.4.0",
+        "react-error-boundary": "^4.0.2",
```
luzzif, line 33: Wondering if we could use an error boundary directly. It's an official React feature and very simple to implement, I don't think we necessarily need a new lib.

https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary

luzzif (changes requested):

packages/react/src/components/error-boundary/index.ts
```diff
@@ -0,0 +1,30 @@
...
+    ErrorBoundaryProps,
+    ErrorBoundaryState
+> {
+    constructor(props: ErrorBoundaryProps) {
+        super(props);
+        this.state = { hasError: false };
+    }
+
+    static getDerivedStateFromError(error: unknown) {
+        console.warn("error catched by boundary", error);
```
luzzif, line 22: Typo, should be "caught", not "catched"

luzzif (dismissed):

luzzif (approved):

PR #209, opened 2023-03-27
feat: first full implementation of ipfs/onchain fetching splitting

luzzif:
This PR splits onchain KPI tokens fetching from IPFS fetching of metadata. This is especially useful for performance as it avoid ONE slow IPFS fetching from blocking the loading of ALL KPI tokens/oracles.


packages/sdk/src/fetcher/core/index.ts
```diff
@@ -209,6 +219,133 @@ class Fetcher implements ICoreFetcher {
...
+    }): Promise<T extends KPIToken ? ResolvedKPIToken : ResolvedOracle> {
+        const template = (
+            await this.resolveTemplates({
+                ipfsGatewayURL,
+                templates: [entity.template],
+            })
+        )[0];
+
+        if (!("oracles" in entity))
+            // FIXME: this is ugly but the ts compiler is complaining otherwise (not sure why)
```
Diogomartf, line 272: I think I'd remove this comment or make it shorter, in my opinion It's better just to have the condition.

packages/sdk/src/fetcher/index.ts
```diff
@@ -15,10 +15,10 @@ import { SubgraphFetcher } from "./subgraph";
 import { CoreFetcher } from "./core";
 import { Token } from "../entities/token";
 
+export * from "./types";
 export * from "./abstraction";
-export * from "./core";
 
-class FullFetcher implements IFullCarrotFetcher {
+class FullFetcher extends CoreFetcher implements IFullCarrotFetcher {
```
Diogomartf, line 21: I don't know if makes sense to extends the class just to get a function used one time.

I think I prefer `CoreFetcher.fetchERC20Tokens({ provider, addresses })`

Also why the class is called _CoreFetcher_, why it's _core_?

Diogomartf:
This two classes share a lot of properties, I think there is an opportunity for a base class with those properties.

image

Diogomartf:
what are the main diferences btw `KPIToken` and `ResolvedKPIToken`?

`ResolvedKPIToken` is a `KPIToken` with a `specification` object?

Does `KPIToken` has  Template and Oracle data?

And what about `ResolvedKPITokenWithData`? Seems like the only difference to `ResolvedKPIToken` is the oracle data.

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -66,16 +68,25 @@ const STATE_OPTIONS = [
...
 
     //  fetch KPITokens
-    const [results, setResults] = useState<KPIToken[]>([]);
-    const { searchQuery, setSearchQuery } = useSearch();
-    const { loading, kpiTokens } = useKPITokens(searchQuery);
+    const [results, setResults] = useState<ResolvedKPIToken[]>([]);
+    const [searchQuery, setSearchQuery] = useState("");
+    const [debouncedSearchQuery, setDebouncedSearchQuery] = useState("");
+
+    useDebounce(() => setDebouncedSearchQuery(searchQuery), 300, [searchQuery]);
```
Diogomartf, line 86: why do you prefer to declare the useDebounce logic explicitly on the component, instead of just using a hook like `useSearch` that hides the underlying logic? 

I think it's much simpler just using this hook.

`const { searchQuery, setSearchQuery } = useSearch();`

And I if I'm working on this component I don't need to care about the debounce for search if not working on it and it's easier to focus if you have less things to care about.

packages/sdk/src/fetcher/index.ts
```diff
@@ -91,21 +89,16 @@ class FullFetcher implements IFullCarrotFetcher {
         return useSubgraph
             ? SubgraphFetcher.fetchKPITokens({
                   provider,
-                  ipfsGatewayURL,
                   addresses,
-                  searchQuery,
```
Diogomartf, line 96: why are we removing the search directly on subgraph?

Diogomartf:
I think this PR shares a lot of things with the PR https://github.com/carrot-kpi/v1-monorepo/pull/166 and it could be better done if we had merged the refactor of `Fetcher` classes first. Now that PR will need a refactor to coupe with the on going changes.

On other note, I think separating the `IPFS` fetching is a good thing and will improve the system. I think it's a gradual step forward like I think the refactor Fetcher classes is.

packages/sdk/src/fetcher/core/index.ts
luzzif: Not sure I'm quite getting it. The comment refers to the explicit cast done a line below.

packages/sdk/src/fetcher/index.ts
luzzif: It makes sense because by doing this we don't have to export the `CoreFetcher` + the overall `Fetcher`.

We can only export the general fetcher that will inherit all of the `CoreFetcher` functionality. It's called `CoreFetcher` because it contains logic that is core to Carrot and decently generic, such as resolving templates, fetching data from IPFS, fetching ERC20 tokens etc etc.

luzzif:
> This two classes share a lot of properties, I think there is an opportunity for a base class with those properties.
> 
> Would also enhance the differences btw them which are hard to spot currently.
> 
> I'm also wondering if it would be better to have them separated by file.
> 
> image

I agree about the possibility for inheritance, I'll try to make it work. As for having them separated by file, I don't think it's necessary.

packages/frontend/src/pages/campaigns/index.tsx
luzzif: It has literally no advantage at all. I mean, declaring a hook that ONLY calls one hook internally rarely makes sense unless there's some things that you want to make transparent to the user. But in this case I think it's overall clearer what this does compared to a `useSearch` (which also has a misleading name, because it has no search-specific logic and is just a hook to return a debounced value).

packages/sdk/src/fetcher/index.ts
luzzif: I think with the separation between fetching and resolving we need to find another way to implement the search functionality. For now with both the subgraph and onchain fetcher we'll do that from the frontend.

luzzif:
> what are the diferences btw `KPIToken` and `ResolvedKPIToken`?
> 
> `ResolvedKPIToken` is a `KPIToken` with a `specification` object?
> 
> Does `KPIToken` has Template and Oracle data?
> 
> And what about `ResolvedKPITokenWithData`? Seems like the only difference to `ResolvedKPIToken` is the oracle data.

So, in general the difference between the base classes and the resolved classes is the fetching of data from IPFS. For example both `KPIToken` and `Oracle` use a `Template` class to represent their template. The `Template` class is a simple representation of the data stored onchain (template id, address, version and sepcification CID), and as such it doesn't contain the resolved template specification, which is instead there in the `ResolvedKPIToken` and `ResolvedOracle` respectively.

As for the `WithData` classes, those represent resolved entities with onchain data returned from the template's `data` function, and are used with template frontends to make sure we're always showing the freshest information possible to the user.

packages/frontend/src/components/ui/kpi-token-card/time-left/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+                {t("stateOptions.finalized")}
+            </Typography>
+        );
+    if (kpiToken.expired)
+        return (
+            <Typography uppercase className={{ root: "text-red" }}>
+                {t("stateOptions.expired")}
+            </Typography>
+        );
+    return <Timer to={kpiToken.expiration} countdown />;
```
guerrap, line 25: Should we also show the seconds here? Or at least expect a `seconds` prop for `TimeLeft` that will be provided to <Timer>?

packages/sdk/src/fetcher/index.ts
Diogomartf: Yes, noticed how it's used and using only fetcher to access this classes is a good thing.

packages/frontend/src/hooks/useSearchedResolvedKPITokens.ts
```diff
@@ -0,0 +1,116 @@
...
+            if (!cancelled) setLoading(false);
+        };
+        void fetchData();
+        return () => {
+            cancelled = true;
+        };
+    }, [ipfsGatewayURL, kpiTokens]);
+
+    useEffect(() => {
+        if (!searchQuery) setSearchedKPITokens(debouncedSearchableKPITokens);
```
guerrap, line 106: We should add an `else` here, or `return` inside the `if` statement, otherwise, the second `setSearchedKPITokens` will always run and overwrite the first one.

packages/react/src/hooks/useFederatedModuleContainer.ts
```diff
@@ -52,24 +52,31 @@ export const useFederatedModuleContainer = (
...
+                        if (!localContainer) {
                             console.warn(
                                 "container still undefined after federated module import"
                             );
                             return;
                         }
-                        await container.init(shareScope);
-                        container.__initialized = true;
-                        if (!cancelled) setContainer(container);
+                        try {
```
guerrap, line 64: I'm not sure if this is the case but this `try/catch` may be "nullifying" the error boundaries around the `TemplateComponent`? Since we are catching the error here, without rethrowing it.

packages/react/src/hooks/useResolvedKPITokenTemplates.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const [resolvedTemplates, setResolvedTemplates] = useState<
+        ResolvedTemplate[]
+    >([]);
+    const [loading, setLoading] = useState(true);
+
+    useEffect(() => {
+        let cancelled = false;
+        async function fetchData(): Promise<void> {
+            if (!chain) return;
```
guerrap, line 25: We can add two more conditions here, like:

```
if(!!ids || ids.length === 0) return;
```

packages/react/src/hooks/useResolvedOracleTemplates.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const [resolvedTemplates, setResolvedTemplates] = useState<
+        ResolvedTemplate[]
+    >([]);
+    const [loading, setLoading] = useState(true);
+
+    useEffect(() => {
+        let cancelled = false;
+        async function fetchData(): Promise<void> {
+            if (!chain) return;
```
guerrap, line 25: See above [here](https://github.com/carrot-kpi/v1-monorepo/pull/209/files#r1150610822)

packages/frontend/src/components/ui/kpi-token-card/time-left/index.tsx
luzzif: The countdown in the card once showed the seconds, but then I removed it because it made the text break and go to the next line, which was ugly. Hence why we now only show minutes.

packages/sdk/src/fetcher/core/index.ts
```diff
@@ -209,6 +219,133 @@ class Fetcher implements ICoreFetcher {
...
+            oracles,
+            JSON.parse(resolvedKPITokenSpecification)
+        );
+    }
+
+    public async resolveKPITokens({
+        ipfsGatewayURL,
+        kpiTokens,
+    }: ResolveKPITokensParams): Promise<ResolvedKPITokensMap> {
+        const resolvedKPITokens = await Promise.all(
```
guerrap, line 314: Are we sure we want to reject if **any** of the token fetch fails?

packages/frontend/src/utils/kpi-tokens.ts
```diff
@@ -1,24 +1,39 @@
-import { KPIToken } from "@carrot-kpi/sdk";
+import { ResolvedKPIToken } from "@carrot-kpi/sdk";
 import { CampaignOrder, CampaignState } from "../pages/campaigns";
 
-export const filterKPITokens = (tokens: KPIToken[], state: CampaignState) => {
+export const filterKPITokens = (
```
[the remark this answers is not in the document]
luzzif: Addressed in d65f11899d21544c1aa81bf671bd18463e535e6c

packages/sdk/src/fetcher/core/index.ts
```diff
@@ -209,6 +219,133 @@ class Fetcher implements ICoreFetcher {
...
+            },
+            {}
+        );
+    }
+
+    public async resolveOracles({
+        ipfsGatewayURL,
+        oracles,
+    }: ResolveOraclesParams): Promise<ResolvedOraclesMap> {
+        const resolvedOracles = await Promise.all(
```
guerrap, line 335: See above.

packages/frontend/src/hooks/useSearchedResolvedKPITokens.ts
luzzif: Addressed in d65f11899d21544c1aa81bf671bd18463e535e6c

packages/frontend/src/utils/kpi-tokens.ts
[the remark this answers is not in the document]
luzzif: Addressed in 90db6d220723602e2a73d6188e5de773a5447bc6

packages/react/src/hooks/useFederatedModuleContainer.ts
luzzif: It shouldn't influence it because we're a step before that. Here we're just initializing the federated module container from which we'll take the React components to use. We'll actually use the component later and the error boundary is applied then.

packages/sdk/src/fetcher/core/index.ts
luzzif: The comment I've put above should also apply here.

packages/sdk/src/fetcher/core/index.ts
luzzif: Should be addressed in c643e9d38303f4dceacedb2dd7f50f8be3aabdda

guerrap (dismissed):

PR #213, opened 2023-03-30
fix(ui): minor style fixes

luzzif:


Diogomartf:
No padding on creation form

image

luzzif:
> No padding on creation form
> 
> image

Wait, where have you taken that screenshot from?

guerrap:
> > > > No padding on creation form
> > > > image
> > > 
> > > 
> > > Wait, where have you taken that screenshot from?
> > 
> > 
> > It's from the preview link, here https://bafybeigqvpgzffi553yamkhrih6mdusij4lssjmg7m2g7kup5pd6qq2g6m.on.fleek.co
> 
> It's pretty weird, but this shouldn't be caused by the PR. The templates are their own piece of code and their style shouldn't be influenced by this.

Ye, exactly, the creation form is working properly on https://carrot-preview.on.fleek.co.

luzzif:
@Diogomartf thanks for the catch, this actually highlighted an issue with the templates, that currently don't bundle UI lib components' CSS. It's a problem because changes in the UI lib styles might cause template styles to break.

This PR addresses the issue: https://github.com/carrot-kpi/erc20-kpi-token-template/pull/92

The plan is to merge this, then do a release of the new template with the bundle styles, and see if the problems still persist in the preview build (they shouldn't).

PR #214, opened 2023-03-30
feat(frontend): add video in hero

luzzif:


Diogomartf:
why do we need so many updates to git command?

luzzif:
> why do we need so many updates to git command?

I was adding git lfs to manage the video in the repo, but now I've just removed it from here and hosted it on an S3 bucket. I've reverted the changes related to git lfs

guerrap (approved):

luzzif:
FYI: the video is now served from our Carrot droplet.

PR #218, opened 2023-04-03
feat(react): implement use erc20 token price hook

luzzif:


packages/react/src/hooks/useERC20TokenPrice.ts
```diff
@@ -0,0 +1,75 @@
...
+        const fetchPrice = async () => {
+            if (!tokenAddress || !isAddress(tokenAddress) || !chain) return;
+            const apiChainPrefix =
+                DEFILLAMA_API_CHAIN_PREFIX[chain.id as ChainId];
+            if (!apiChainPrefix) return;
+            if (!cancelled) setLoading(true);
+            try {
+                const checksummedAddress = getAddress(tokenAddress);
+                const keys = [`coingecko:${checksummedAddress}`];
+                if (apiChainPrefix)
```
guerrap, line 33: I think we can remove this check here right? We are already checking for its existence before in:
```ts
if (!apiChainPrefix) return;
```
Here it would always exist.

luzzif: Good catch, but we actually need to remove the first check. Will do that.

luzzif: Addressed in 78915bacf99b4945f1fa07e5ec22fdb18f8aca60

packages/react/src/hooks/useERC20TokenPrice.ts
```diff
@@ -0,0 +1,75 @@
...
+                const body = (await response.json()) as {
+                    coins: Record<
+                        string,
+                        { price: number; confidence: number }
+                    >;
+                };
+
+                for (const key of keys) {
+                    const wrappedPrice = body.coins[key];
+                    console.log({ body, key, wrappedPrice });
```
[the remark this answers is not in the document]
luzzif: Addressed in c77873f9b8860d325a8383390bef8f2a092883e1

PR #219, opened 2023-04-04
feat: add pagination

Diogomartf:
**summary**
This PR adds frontend pagination to campaigns page

**Disclaimer ⚠️** 
As we, currently, have small dataset, I put the items per page as 1 to be easy to test. I'll revert it before merging.

**Screenshots example**
_The items per page is set to 1 for demonstration purposes._

```
items_per_page = 1
total_pages = 5
```

 
1. 
image
2.
image
3.
image

4.
image

5.
image


packages/react/src/hooks/usePagination.ts
```diff
@@ -0,0 +1,16 @@
+const ITEMS_PER_PAGE = 1;
```
Diogomartf, line 1: this is the default items per page, will be updated before merge.

MilanVojnovic95:
Tested and looks ok to me 👍

packages/frontend/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,122 @@
+import React, { Dispatch, SetStateAction } from "react";
+import { PaginationBall } from "./pagination-ball";
+import { ReactComponent as ArrowIcon } from "../../assets/arrow.svg";
+import { cva } from "class-variance-authority";
+
+const createArrayBetweenRange = (start: number, end: number) =>
```
guerrap, line 6: I think that we can remove this function, since it's only used one time.

packages/frontend/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    Array.from({ length: end - start }, (_, i) => start + i);
+
+const LOT_OF_PAGES = 5;
```
guerrap, line 9: This could be renamed to something like `VISIBLE_PAGE_NAVIGATION_THRESHOLD`.

packages/frontend/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+    totalPages: number;
+    currentPage: number;
+    setCurrentPage: Dispatch<SetStateAction<number>>;
+}) => {
+    const isFirstPage = currentPage === 1;
+    const isLastPage = currentPage === totalPages;
+    const hasMoreThanOnePage = totalPages > 1;
+    const hasALotOfPages = totalPages > LOT_OF_PAGES;
+    const isNotFirstPage = currentPage > 1;
+    const isNotLastPage = currentPage < totalPages;
```
guerrap, line 33: We can remove `isNotFirstPage` and `isNotFirstPage` here, and use `isFirstPage` and `isLastPage` with a negation `!`, that should work.

I think we can remove `hasALotOfPages` too, and just use the const value when needed.

Also, I'd rename these constants like:

```tsx
    const firstPage = currentPage === 1;
    const lastPage = currentPage === totalPages;
    const moreThanOnePage = totalPages > 1;
```

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -35,16 +39,28 @@ export const Campaigns = () => {
...
-
     // page settings
-    const [pageSize, setPageSize] = useState(12);
+    const [filtersOpen, setFilterOpen] = useState(false);
+    const toggleFilters = () => setFilterOpen(!filtersOpen);
     const [ordering, setOrdering] = useState<SelectOption>(ORDERING_OPTIONS[0]);
+    const [currentPage, setCurrentPage] = useState(1);
+    const { data, totalPages } = usePagination(results, currentPage, 4);
+
+    const handlePageChange = (pageNumber: SetStateAction<number>) => {
```
guerrap, line 49: We should wrap this with a `useCallback` since the function gets passed down to `Pagination` and then `PaginationBall`.

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -186,8 +177,13 @@ export const Campaigns = () => {
                                                   );
                                               })}
                                     </div>
-                                    <Pagination />
-                                </>
+
+                                    <Pagination
```
guerrap, line 181: The `Pagination` shouldn't be visible if the KPI tokens are still loading:

```tsx
{!loading && data.length > 0 && <Pagination />}
```

packages/frontend/src/components/pagination/pagination-ball/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+    currentPage,
+    children,
+}: {
+    number?: number;
+    onClick?: MouseEventHandler<HTMLDivElement>;
+    currentPage?: number;
+    children?: ReactNode;
+}) => {
+    const isActive = Number(number) === currentPage;
+    const isClickable = !!onClick;
```
guerrap, line 31: I'd remove these two `const` and put the conditions directly in the `paginationBallStyles`.

packages/frontend/src/components/pagination/pagination-ball/index.tsx
```diff
@@ -0,0 +1,40 @@
+import { cva } from "class-variance-authority";
+import React, { MouseEventHandler, ReactNode } from "react";
+
+const paginationBallStyles = cva(
+    ["flex items-center justify-center", " w-12 h-12", "rounded-full"],
```
guerrap, line 5: Typo here, there's a space `" w-12 h-12"`.

guerrap:
The pagination should be aligned to the left, following Figma:
image

packages/frontend/src/components/pagination/index.tsx
Diogomartf: In this case as it a bit of verbose code I do prefer to encapsulate in a function. 

Do you think we can left it like this or do you have a strong opinion on that?

packages/frontend/src/components/pagination/pagination-ball/index.tsx
Diogomartf: thanks

packages/frontend/src/components/pagination/index.tsx
Diogomartf: I do find it more difficult to understand than a "LOT_OF_PAGES" 😅

Diogomartf: Maybe `VISIBLE_PAGES_THRESHOLD`, I'd remove navigation.

packages/frontend/src/components/pagination/index.tsx
Diogomartf: True. Much better. Thanks.

I'll keep the hasALotOfPages variable cause it's simpler than doing the condition and also helps to explain the VISIBLE_PAGES_THRESHOLD variable.

```js
    const hasALotOfPages = totalPages > VISIBLE_PAGES_THRESHOLD;
```

packages/frontend/src/components/pagination/index.tsx
Diogomartf: I fixed to VISIBLE_PAGES_THRESHOLD. thanks

packages/frontend/src/components/pagination/pagination-ball/index.tsx
Diogomartf: Yes, that might work better. thanks

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: True. Thanks!

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: I'll add it. Thanks

packages/frontend/src/components/pagination/index.tsx
Diogomartf: fixed: fdd6ea5450d63dad6d1fd228a8acafda08ba2e07

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: fixed: c3f356ab938dc7946f877253851d47318df5df6b

packages/frontend/src/pages/campaigns/index.tsx
Diogomartf: fixed: cf4ac14e2f9252827bc786c3e9b4f006b3c884cb

packages/frontend/src/components/pagination/pagination-ball/index.tsx
Diogomartf: fixed: d404c18e7f424abd4115012d89d1e801d240a720

packages/frontend/src/components/pagination/pagination-ball/index.tsx
Diogomartf: fixed: d404c18e7f424abd4115012d89d1e801d240a720

Diogomartf:
> The pagination should be aligned to the left, following Figma: image

yeah, I know, but with current alignment of cards I do think it's better to have it aligned center.

image

aligning it at the left will make it worse (IMO)

image

---

We have several things that are not aligned. I think center is better played right now.

packages/frontend/src/components/pagination/index.tsx
guerrap: Let's leave it like this for now.

guerrap (dismissed):

guerrap (approved):

luzzif (changes requested):

packages/frontend/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+            false: "invisible",
+        },
+    },
+});
+
+export const Pagination = ({
+    totalPages,
+    setCurrentPage,
+    currentPage,
+}: {
```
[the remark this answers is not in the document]
Diogomartf: yup, true, I'll change it.

packages/frontend/src/components/pagination/pagination-ball/index.tsx
```diff
@@ -0,0 +1,39 @@
+import { cva } from "class-variance-authority";
+import React, { MouseEventHandler, ReactNode } from "react";
+
+const paginationBallStyles = cva(
+    ["flex items-center justify-center", "w-12 h-12", "rounded-full"],
+    {
+        variants: {
+            isActive: {
```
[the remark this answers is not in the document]
Diogomartf: Didn't know that was a pattern in java, I generally use it when a variable is a boolean.

We can do that only on functions.

I'll fix this.

packages/frontend/src/components/pagination/pagination-ball/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+        },
+    }
+);
+
+export const PaginationBall = ({
+    number,
+    onClick,
+    currentPage,
+    children,
+}: {
```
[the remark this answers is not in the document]
Diogomartf: fixed: 54fa95a1c9a3d7d968bd26f2585236aa6eb6ae79

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -35,16 +40,54 @@ export const Campaigns = () => {
...
-        useSearchedResolvedKPITokens(debouncedSearchQuery);
-
-    //  toggle filters
-    const toggleFilters = () => setFilterOpen(!filtersOpen);
-    const [filtersOpen, setFilterOpen] = useState(false);
-
     // page settings
-    const [pageSize, setPageSize] = useState(12);
+    const [filtersOpen, setFilterOpen] = useState(false);
+    const toggleFilters = () => setFilterOpen(!filtersOpen);
```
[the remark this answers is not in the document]
Diogomartf: fixed: d11b7fdc3c60d734e3fddea837dc881b7465793f

packages/frontend/src/components/pagination/pagination-ball/index.tsx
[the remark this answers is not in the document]
Diogomartf: fixed: 54fa95a1c9a3d7d968bd26f2585236aa6eb6ae79

packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -102,31 +145,6 @@ export const Campaigns = () => {
         setResults(sortedFilteredTokens);
     }, [sortedFilteredTokens]);
 
-    const page = usePage(results, pageSize, 0);
-
-    const resizeObserver = useRef(
```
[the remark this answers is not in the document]
luzzif: It's to show more or less cards based on the screen size. For example while consuming the dapp from the desktop, up to 12 cards could be shown in a page (in a 4x3 format) but while consuming the dapp on mobile, the cards are all in a single column, and having 12 of them means the user would have to scroll quite a bit to change page. That's why the resize observer was there. On mobile we showed like 3 cards, scaling the amount up until desktop size was reached.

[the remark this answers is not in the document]
guerrap: I didn't notice this change; having the page size change on the screen size seems reasonable.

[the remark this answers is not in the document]
Diogomartf: I actually don't agree with this approach. 

It's a lot of code to do something I don't think we need.

I don't think it's a problem to have 18 cards on mobile, it's not that many and users are there to see the campaigns cards, there's not anything bellow the cards to do.

That's also what most sites show on lists, for eg. Airbnb (18 cards), OpenSea (a lot more than 18). They also show the same cards on mobile and desktop.

I would prefer to keep the code as it is (removing this extra code) and if we think it's a problem in the future we can always change it.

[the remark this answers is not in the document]
guerrap: Idk at this point we can try to remove that piece of code and see how it goes; another possible solution would be to have the campaigns scroll horizontally on mobile maybe?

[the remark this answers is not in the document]
Diogomartf: It's easier to scroll vertical than horizontally. So I'd keep it vertical.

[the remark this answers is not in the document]
luzzif: Let's change this. Full stop.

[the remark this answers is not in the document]
Diogomartf: Added back ResizeObserver.

guerrap (approved):

luzzif (approved):

PR #224, opened 2023-04-06
fix(frontend): fix service worker by using workbox to cache ipfs requests

luzzif:


Diogomartf (approved):

packages/frontend/craco.config.js
```diff
@@ -40,28 +36,14 @@ module.exports = {
...
-                chunkFilename: "[name].[contenthash:8].js",
-                assetModuleFilename: "[name].[contenthash:8][ext]",
-            };
-            // TODO: make Workbox and precaching work
-            // config.plugins.push(
-            //     new WorkboxWebpackPlugin.InjectManifest({
-            //         swSrc: join(__dirname, "/src/sw.ts"),
-            //     })
-            // );
+            // TODO: make Workbox and precaching work (all
```
guerrap, line 39: We can remove the TODO since we are using `Workbox` now, right?

guerrap (approved):

packages/frontend/craco.config.js
luzzif: Nono, right now we're basically only using that plugin to correctly transpile the service worker, but we're not really leveraging Workbox (that will be done when we enable precaching and turn Carrot into a proper PWA).

PR #226, opened 2023-04-06
fix(ui): markdown

luzzif:


Diogomartf:
Markdown not picking font family 

image

packages/ui/src/utils/components/index.tsx
```diff
@@ -29,3 +29,10 @@ export function mergedCva<T extends object>(
         return twMergeWithConfig(classList(...classListParams));
     };
 }
+
+export function mergedCx(
```
Diogomartf, line 33: not sure about the name of this function 🤔

luzzif: It's the `cx` function with TW merge applied. It's following the same name conventions as `mergedCva`

luzzif:
> Markdown not picking font family
> 
> image

I'm actually seeing it right now, can you retest?

Diogomartf (approved):

PR #227, opened 2023-04-07
fix: make video pause when opening modal, also remove autoplay

Diogomartf:
**summary**

The video should autoplay on mute. 

If a modal is open, it should pause and when the modal is closed it should play again.

image

---

**Test flow**

1. Go to landing
2. Check if the video is playing on mute
3. open a modal (video should pause)
4. close modal (video should play)

### DIFF
--- a/packages/frontend/src/components/fullscreen-modal/index.tsx
+++ b/packages/frontend/src/components/fullscreen-modal/index.tsx
@@ -1,7 +1,9 @@
-import React, { ReactNode } from "react";
+import React, { ReactNode, useEffect } from "react";
 import { Navbar, NavbarProps } from "../ui/navbar";
 import { animated, SpringValue } from "@react-spring/web";
 import { cva } from "class-variance-authority";
+import { setModalIsOpen } from "../../state/reducers/modals";
+import { useDispatch } from "../../state/connector";
 
 const rootStyles = cva(
     [
@@ -36,9 +38,19 @@ export const AnimatedFullscreenModal = ({
     onDismiss,
     children,
 }: FullscreenModalProps) => {
+    const dispatch = useDispatch();
+
+    useEffect(() => {
+        dispatch(setModalIsOpen({ isOpen: true }));
+
+        return () => {
+            dispatch(setModalIsOpen({ isOpen: false }));
+        };
+    }, [dispatch]);
+
     return (
         <animated.div style={springStyle} className={rootStyles({ bgColor })}>
-            <div className="flex flex-col h-full w-full">
+            <div className="flex flex-col w-full h-full">
                 <Navbar mode="modal" bgColor={bgColor} onDismiss={onDismiss} />
                 <div className="flex-grow">{children}</div>
             </div>
--- a/packages/frontend/src/components/fullscreen-modal/index.tsx
+++ b/packages/frontend/src/components/fullscreen-modal/index.tsx
@@ -41,10 +41,10 @@ export const AnimatedFullscreenModal = ({
     const dispatch = useDispatch();
 
     useEffect(() => {
-        dispatch(setModalOpen({ open: true }));
+        dispatch(setModalOpen(true));
 
         return () => {
-            dispatch(setModalOpen({ open: false }));
+            dispatch(setModalOpen(false));
         };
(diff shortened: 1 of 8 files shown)


packages/frontend/src/pages/home/hero/index.tsx
```diff
@@ -28,25 +30,24 @@ export const Hero = () => {
...
+    const modalIsOpen = useSelector<HostState>((state) => state.modals.isOpen);
 
     useEffect(() => {
         if (!videoRef || !videoRef.current) return;
         const currentRef = videoRef.current;
-        return () => {
-            currentRef.pause();
-        };
-    });
+        modalIsOpen ? currentRef.pause() : currentRef.play();
```
guerrap, line 38: I'd avoid playing the video when the modal closes, otherwise every time the user opens and closes a model the video starts even if it was previously over.
I'd just stop it if any modal gets opened but then leave the choice to the user to start the video manually.

packages/frontend/src/state/reducers/modals/types/index.ts
```diff
@@ -0,0 +1,3 @@
+export interface ModalsState {
+    isOpen: boolean;
```
guerrap, line 2: I'd rename this to just `open`.

packages/frontend/src/state/reducers/modals/actions/index.ts
```diff
@@ -0,0 +1,5 @@
+import { createAction } from "@reduxjs/toolkit";
+
+export const setModalIsOpen = createAction<{
```
guerrap, line 3: I'd rename this `setModalOpen`.

packages/frontend/src/pages/home/hero/index.tsx
guerrap, line 38: I'd avoid playing the video when the modal closes, otherwise every time the user opens and closes a model the video starts even if it was previously over.
I'd just stop it if any modal gets opened but then leave the choice to the user to start the video manually.

guerrap (changes requested):

packages/frontend/src/pages/home/hero/index.tsx
Diogomartf: Yes, that might be better.

Diogomartf: fixed 2059e7e1c1e6215fab40c71b46b81713a95385a8

packages/frontend/src/state/reducers/modals/types/index.ts
Diogomartf: fixed: 133bb15676195f8a8a2cdb417ac0dc5fc4172360

packages/frontend/src/state/reducers/modals/actions/index.ts
Diogomartf: fixed: 133bb15676195f8a8a2cdb417ac0dc5fc4172360

guerrap (dismissed):

packages/frontend/src/state/reducers/modals/actions/index.ts
```diff
@@ -0,0 +1,5 @@
+import { createAction } from "@reduxjs/toolkit";
+
+export const setModalOpen = createAction<{
```
luzzif, line 3: There will always be one payload attribute and it will be the open state, so we can simplify and rewrite this as:

```
export const setModalOpen = createAction<boolean>("modals/setModalOpen");
```

packages/frontend/src/components/fullscreen-modal/index.tsx
```diff
@@ -36,9 +38,19 @@ export const AnimatedFullscreenModal = ({
     onDismiss,
     children,
 }: FullscreenModalProps) => {
+    const dispatch = useDispatch();
+
+    useEffect(() => {
+        dispatch(setModalOpen({ open: true }));
```
luzzif, line 44: With the suggested change below, this would become `setModalOpen(true)` and `setModalOpen(false)`

luzzif (changes requested):

guerrap (approved):

PR #230, opened 2023-04-12
fix: not keeping same network on refresh

### LINKED ISSUE #171: refreshing the page loses the chosen network
image

How to reproduce:
- open carrot
- change the network
- refresh
- check that it went back to default network

The intended way is to keep the users choice.

Diogomartf:
closes #171

**Summary**
After trying a couple of things, I found that chain id is present on wagmi store on local storage so we can get the previous chain id from there.

Using the `this.getChainId()` method was not working because its using the provider and throws an error and was going to catch clause

```js
} catch (e) {
    useDefaultChainId = true;
}
```

which set `useDefaultChainId=true` and always returned our default chain id, gnosis.


packages/frontend/src/connectors/readonly/index.ts
```diff
@@ -2,14 +2,33 @@ import { JsonRpcProvider } from "@ethersproject/providers";
...
 
 export const READ_ONLY_CONNECTOR_ID = "readonly";
+const DEFAULT_CHAIN_ID = ChainId.GNOSIS; // 100
 
 export type InjectedConnectorOptions = {
     /** Name of connector */
     name?: string | ((detectedName: string | string[]) => string);
 };
 
+const getChainIdFromlocalStorage = () => {
```
guerrap, line 15: I think we should move this function to `utils/state`.

packages/frontend/src/connectors/readonly/index.ts
```diff
@@ -2,14 +2,33 @@ import { JsonRpcProvider } from "@ethersproject/providers";
 import { Connector } from "wagmi";
 import { Address, Chain, ConnectorData, normalizeChainId } from "@wagmi/core";
 import { constants } from "ethers";
+import { ChainId } from "@carrot-kpi/sdk";
 
 export const READ_ONLY_CONNECTOR_ID = "readonly";
+const DEFAULT_CHAIN_ID = ChainId.GNOSIS; // 100
```
guerrap, line 8: We can remove the comment `// 100` here.

packages/frontend/src/connectors/readonly/index.ts
```diff
@@ -2,14 +2,33 @@ import { JsonRpcProvider } from "@ethersproject/providers";
...
 
 export type InjectedConnectorOptions = {
     /** Name of connector */
     name?: string | ((detectedName: string | string[]) => string);
 };
 
+const getChainIdFromlocalStorage = () => {
+    const wagmiStore = localStorage.getItem("wagmi.store");
+    if (!wagmiStore) return DEFAULT_CHAIN_ID;
+    const parsedStore = JSON.parse(wagmiStore);
```
guerrap, line 18: Should we add a type for the parsed wagmi state object?

packages/frontend/src/connectors/readonly/index.ts
Diogomartf: Yes.

packages/frontend/src/connectors/readonly/index.ts
Diogomartf: added `wagmiStoreObj` interface

Diogomartf: added `wagmiStoreObj` interface

packages/frontend/src/connectors/readonly/index.ts
Diogomartf: fixed 43c7045c51411769a757cd05286ca9fc549dcc23

packages/frontend/src/connectors/readonly/index.ts
Diogomartf: fixed on e0b89de42772c6efefdb918a999147a0a12a0c4f

packages/frontend/src/connectors/readonly/index.ts
```diff
@@ -2,14 +2,43 @@ import { JsonRpcProvider } from "@ethersproject/providers";
...
 
 export const READ_ONLY_CONNECTOR_ID = "readonly";
+const DEFAULT_CHAIN_ID = ChainId.GNOSIS;
 
 export type InjectedConnectorOptions = {
     /** Name of connector */
     name?: string | ((detectedName: string | string[]) => string);
 };
 
+interface wagmiStoreObj {
```
guerrap, line 15: We can change this to `WagmiStore` (also with capital `W`).
I think `state, data, chain, id` should all be optional with `?`, since they may be `undefined`.

Diogomartf: fixed: bce2c26da9eea0de96637b3dfa30bee94e97c732

guerrap (approved):

guerrap (dismissed):

packages/frontend/src/connectors/readonly/index.ts
```diff
@@ -2,14 +2,43 @@ import { JsonRpcProvider } from "@ethersproject/providers";
...
 
 export const READ_ONLY_CONNECTOR_ID = "readonly";
+const DEFAULT_CHAIN_ID = ChainId.GNOSIS;
 
 export type InjectedConnectorOptions = {
     /** Name of connector */
     name?: string | ((detectedName: string | string[]) => string);
 };
 
+interface WagmiStore {
```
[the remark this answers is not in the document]
Diogomartf: To that will need the code to save chainID to local storage and keep that data correct (every time the network changes).

Wagmi give us that right now, the risk is the lib changing but I would embrace that risk for now.

packages/frontend/src/connectors/readonly/index.ts
```diff
@@ -119,6 +137,7 @@ export class ReadonlyConnector extends Connector<
     protected onAccountsChanged = (): void => {};
 
     protected onChainChanged = (chainId: number | string): void => {
+        console.log("chainId:", chainId);
```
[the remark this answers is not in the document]
Diogomartf: fixed.

MilanVojnovic95:
Tested and looks ok to me 👍

packages/frontend/src/connectors/readonly/index.ts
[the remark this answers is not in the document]
luzzif: The code for doing the above is like 20 lines top... We just need to add the current `chainId` to the state, have an action to set it, and have an updater with a `useEffect` to intercept chain changes and call the action...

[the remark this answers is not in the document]
Diogomartf: I can try that.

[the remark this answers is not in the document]
Diogomartf: fixed here: https://github.com/carrot-kpi/v1-monorepo/pull/230/commits/7a034540e650c691022b0e371a63f7e99af6b9d2

packages/frontend/src/connectors/readonly/index.ts
```diff
@@ -24,24 +25,18 @@ export class ReadonlyConnector extends Connector<
         super(config);
     }
 
+    public getChainIdFromState() {
```
luzzif, line 28: Why is this function a function? Would we ever need to access the chain id saved in the store from somewhere? And especially by accessing this connector?

luzzif (changes requested):

packages/frontend/src/components/connect-wallet/index.tsx
```diff
@@ -77,12 +78,13 @@ export const ConnectWallet = () => {
         async (chainId: number) => {
             try {
```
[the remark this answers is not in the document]
Diogomartf: I didn't thought about hat use case, will take a look.

packages/frontend/src/connectors/readonly/index.ts
Diogomartf: You are right, doesn't need to be on the class, I guess we can use it directly for now.

Diogomartf: fixed: https://github.com/carrot-kpi/v1-monorepo/pull/230/commits/32a0363aa3f971977a3dd77c1a7b5940fd1651fc

packages/frontend/src/components/connect-wallet/index.tsx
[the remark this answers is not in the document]
Diogomartf: We can have an effect that listens to the chain, and when it changes sets the new chainId in the store.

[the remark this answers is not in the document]
Diogomartf: [check this solution](https://github.com/carrot-kpi/v1-monorepo/pull/230/commits/615593b4686a3ef9f39a5348c9711c3fcac70f8a)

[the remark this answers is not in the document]
Diogomartf: I don't know if I understand the `defaultChainId`, and where should we use `chainId` vs `defaultChainId`

PR #233, opened 2023-04-18
Mobile compatibility

guerrap:
The `DatePicker`  component was causing weird horizontal scrolling on mobile devices since the picker contained by the popover is already on the DOM (with 0 opacity).
Having the `DatePicker` popover as a modal on mobile solves the issue.

### DIFF
--- a/packages/ui/src/components/input/date-time/index.tsx
+++ b/packages/ui/src/components/input/date-time/index.tsx
@@ -46,14 +46,6 @@ export const DateTimeInput = forwardRef<HTMLInputElement, DateTimeInputProps>(
         const [open, setOpen] = useState(false);
         const [modal, setModal] = useState(false);
 
-        const resizeObserver = useRef(
-            new ResizeObserver((entries) => {
-                const { width } = entries[0].contentRect;
-                if (width < 640) setModal(true);
-                else setModal(false);
-            })
-        );
-
         const resolvedId = id || generatedId;
 
         useClickAway(popoverRef, () => {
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/input/date-time/index.tsx
```diff
@@ -37,16 +44,36 @@ export const DateTimeInput = forwardRef<HTMLInputElement, DateTimeInputProps>(
         const [anchorEl, setAnchorEl] = useState<HTMLInputElement | null>(null);
         const popoverRef = useRef<HTMLDivElement>(null);
         const [open, setOpen] = useState(false);
+        const [modal, setModal] = useState(false);
+
+        const resizeObserver = useRef(
```
luzzif, line 49: Not sure a ref is necessary here. If we declare the `ResizeObserver` in the `useEffect` hook below directly it should be the same (minus the ref we wouldn't use) as it's not used anywhere else.

packages/ui/src/components/input/date-time/index.tsx
```diff
@@ -37,16 +44,36 @@ export const DateTimeInput = forwardRef<HTMLInputElement, DateTimeInputProps>(
...
+            return () => {
+                // eslint-disable-next-line react-hooks/exhaustive-deps
+                resizeObserver.current.unobserve(document.body);
+            };
+        }, []);
+
         const handlePickerOpen = useCallback(() => {
             setOpen(true);
         }, []);
+        const handlePickerClose = useCallback(() => {
```
luzzif, line 74: Missing new line here

packages/ui/src/components/input/date/index.tsx
```diff
@@ -37,16 +44,36 @@ export const DateInput = forwardRef<HTMLInputElement, DateInputProps>(
...
+            return () => {
+                // eslint-disable-next-line react-hooks/exhaustive-deps
+                resizeObserver.current.unobserve(document.body);
+            };
+        });
+
         const handlePickerOpen = useCallback(() => {
             setOpen(true);
         }, []);
+        const handlePickerClose = useCallback(() => {
```
luzzif, line 74: Ditto here

luzzif (changes requested):

packages/ui/src/components/input/date/index.tsx
```diff
@@ -37,16 +44,36 @@ export const DateInput = forwardRef<HTMLInputElement, DateInputProps>(
...
             setOpen(false);
         });
 
+        useEffect(() => {
+            resizeObserver.current.observe(document.body);
+            return () => {
+                // eslint-disable-next-line react-hooks/exhaustive-deps
+                resizeObserver.current.unobserve(document.body);
+            };
+        });
```
Diogomartf, line 69: this useEffect is the same as above, as it is the `resizeObserver = useRef`

Maybe there is an opportunity to extract it.

guerrap: I'd leave it like this for now, since it's just a single `useEffect`.

luzzif (approved):

PR #234, opened 2023-04-18
fix: campaigns now shows the loading correctly

### LINKED ISSUE #223: All campaigns loading issue
On the campaigns page, when the loading is completed there's an issue that causes the `<Empty />` component to be rendered for a second, before showing the actual campaigns.

This is probably caused by the `useDebounce` inside the `useSearchedResolvedKPITokens`, and how that hook is Implemented.

https://user-images.githubusercontent.com/9011637/230117599-f14e7ad6-7b08-4fda-bfd5-00cc86f3f807.mp4

Diogomartf:
closes #223


packages/frontend/src/hooks/search/filters/index.ts
```diff
@@ -0,0 +1,32 @@
+import { ResolvedKPIToken } from "@carrot-kpi/sdk";
+
+export const filterTokensWithQuery = (
```
luzzif, line 3: This function is pretty pointless imho

luzzif (changes requested):

packages/frontend/src/hooks/search/filters/index.ts
Diogomartf: this is a nice abstraction, otherwise will have to do the looping on the useEffect


```js
const searchedTokens = filterTokensWithQuery(
    searchQuery,
    initialTokens
);
```

I thinks nicer this way.

packages/frontend/src/hooks/search/filters/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    query: string,
+    tokens: ResolvedKPIToken[]
+) => {
+    return tokens.filter((token) =>
+        tokenSpecificationIncludesQuery(token, query)
+    );
+};
+
+export const tokenSpecificationIncludesQuery = (
```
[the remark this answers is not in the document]
Diogomartf: same, answer, I think it's better organised this way.

I can remove the export as it isn't needed elsewhere.

packages/frontend/src/hooks/search/filters/index.ts
```diff
@@ -0,0 +1,32 @@
+import { ResolvedKPIToken } from "@carrot-kpi/sdk";
```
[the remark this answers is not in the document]
Diogomartf: Yeah. I wondered where to put it, as it used in a hook I wanted to put it close. So created a folder for search on hooks folder with the `hook` and `search utils`

But yeah, I can move it to utils folder.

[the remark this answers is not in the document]
Diogomartf: Moved to utils.

fixed here: 4767926cbfa674bb6cbb228e812befca257d4bc2

MilanVojnovic95:
Tested and it’s the same as it was before, for a second we can se empty campaign cards, and also when u click on "View All Campaigns" it sends you at the bottom of the page so we must scroll up.

Diogomartf:
> Tested and it’s the same as it was before, for a second we can se empty campaign cards, and also when u click on "View All Campaigns" it sends you at the bottom of the page so we must scroll up.

The "View All Campaigns" button is on landing page, this is meant to fix "campaigns" page.
Can you confirm if is working on campaigns page, tried in different devices and all seem to work.

I can fix that scroll on click in other PR.

packages/frontend/src/hooks/search/filters/index.ts
luzzif: No but ser, it's literally pointless. What use have we for an additional function call (overhead) when the function is just a single instruction... there's no need to define one-line functions that are not called anywhere else, it doesn't add in clarity nor tidiness to the code (in fact, it removes it). If you want to make it clearer what's going on a comment is a better option.

Diogomartf: Ok, I buy this. 

I'll remove the abstraction.

Diogomartf: Have to agree, it's better now.

fixed: [6f8bc30](https://github.com/carrot-kpi/v1-monorepo/pull/234/commits/6f8bc30b9bcd66a3086016f2ef496fe641ff7286)

guerrap (approved):

luzzif (approved):

PR #235, opened 2023-04-21
Fathom

guerrap:
Add Fathom tracking for page navigation.
Add integration with the Fathom APIs to automatically create any new events to the remote site.


Diogomartf:
Do you have a link for the Fathom dashboard?

Diogomartf (dismissed):

guerrap:
> Do you have a link for the Fathom dashboard?

Yup, here https://app.usefathom.com/share/ypuqowiv/carrot.

packages/frontend/src/hooks/useAddTransaction.ts
```diff
@@ -20,5 +21,7 @@ export const useAddTransaction = () => {
                 serializedTx: serializeTransaction(tx),
             })
         );
+        if (!fathom) return;
+        fathom.trackRegisteredGoal(TX_FATHOM_EVENTS[tx.type], 0);
```
luzzif, line 25: Isn't this operation async? Maybe it's worth it to encapsulate this in a try catch block

packages/frontend/src/hooks/useFathomTrackPageWatch.ts
```diff
@@ -0,0 +1,13 @@
...
+import { useEffect } from "react";
+import { useLocation } from "react-router-dom";
+
+export const useFathomTrackPageWatch = () => {
+    const location = useLocation();
+    const fathom = useFathom();
+
+    useEffect(() => {
+        if (!fathom) return;
+        fathom.trackPageview({ url: location.pathname });
```
luzzif, line 11: Does this react to a network switch? Do we want to track network switches too?

packages/react/bin/sync-fathom-events/index.ts
```diff
@@ -0,0 +1,63 @@
+import { writeFile } from "fs/promises";
```
luzzif, line 1: It would be cool to have this in a `scripts` folder. This is clearly not a binary.

packages/react/src/analytics/generated.ts
```diff
@@ -0,0 +1,71 @@
+/*
```
luzzif, line 1: Does it make sense to check this into git since it's autogenerated?

packages/react/src/analytics/generated.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export interface FathomSiteInformation {
+    siteId: string;
+    events: FathomSiteEvent[];
+    timestamp: number;
+}
+
+export const siteEvents: FathomSiteInformation = {
+    siteId: "YPUQOWIV",
```
luzzif, line 26: This should probably be an env

packages/react/src/analytics/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./analytics";
```
luzzif, line 1: We should probably move away from the `folder-name/index.ts` naming convention. I thought it had its advantages but I far overstimated them. It's probably better to avoid the folder/index structure and name files using kebab case. In case a module needs to be split up into multiple submodules it still makes sense to have a folder with an index file, but this doesn't seem to be the case.

luzzif (changes requested):

packages/react/bin/sync-fathom-events/index.ts
guerrap: Fixed in bba31bed15c202f8cfeee7cafeaf0a70ef7395d6.

packages/frontend/src/hooks/useFathomTrackPageWatch.ts
guerrap: No, currently this does not react to the network switch since it's just the query param that changes and not the actual path; anyway I don't know if it makes sense to track network switches (for now, maybe in the future).

packages/react/src/analytics/index.ts
guerrap: Fixed in 4b248304b183e32434a4dcc4198173a3cfe055a3.

packages/react/src/analytics/generated.ts
guerrap: Fixed in 43032df6185e2fc20d89b2de720e23e45d81d7f6.

luzzif (changes requested):

packages/react/scripts/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
guerrap: Thinking about it, we can just ignore the `scripts` from the typescript build since the script is run using `ts-node` (also trying to use a single tsconfig with multiple routes caused annoying issues with the exported types).

packages/react/src/analytics/index.ts
luzzif: On this I wrote that it makes sense if we have a module with multiple submodules needed for clarity. It seems like this is the case. In this scenario it made sense to have a folder with the index exporting the 2 submodules, sorry.

packages/frontend/src/hooks/useAddTransaction.ts
luzzif: Maybe just a warn log is enough, it's not something that is really stopping the app from working

guerrap: Fixed In c2a0c120e14abab42b64aea69a86403598289f14.

packages/react/src/analytics/index.ts
guerrap: Fixed in 6490196e9b0b770b39a522af6265e9917ec2a327.

luzzif (approved):

PR #238, opened 2023-04-26
feat(ui, frontend): upgrade dependencies and more

luzzif:
- Upgrades Storybook to v7
- Upgrades other dependencies which are safe to upgrade
- Converts SVGs to JSX components (allows to remove SVGR and URL loader Webpack and Rollup plugins and overall makes configs simpler)
- Removes explicit setting of `line-clamp` plugin since Tailwind v3.3 has moved the plugin to the core release
- Improves KPI token template and campaign cards making their description section scrollable instead of clamping


packages/frontend/src/components/ui/kpi-token-card/index.tsx
```diff
@@ -72,23 +72,22 @@ export const KPITokenCard = ({ kpiToken, noBorder }: KPITokenCardProps) => {
...
-                            <div className="flex flex-col gap-3">
-                                <Skeleton width="100%" />
-                                <Skeleton width="100%" />
-                                <Skeleton width="100%" />
-                                <Skeleton width="20%" />
-                            </div>
-                        )}
+                    <div className="relative">
+                        <div className="absolute pointer-events-none top-0 left-0 w-full h-full shadow-vertical-scroller shadow-white dark:shadow-black" />
+                        <div className="h-40 p-4 overflow-y-auto scrollbar-none ">
```
guerrap, line 77: Empty space here.

packages/frontend/src/constants/index.ts
```diff
@@ -2,18 +2,22 @@ import { ChainId } from "@carrot-kpi/sdk";
...
 
 export interface AugmentedChain extends Chain {
-    logo: FunctionComponent<SVGProps<SVGSVGElement>>;
+    logo: FunctionComponent<
+        SVGProps<SVGSVGElement> & { title?: string | undefined }
+    >;
     iconBackgroundColor: string;
 }
 
+// FunctionComponent<SVGProps<SVGSVGElement> & { title?: string | undefined; }>
```
guerrap, line 19: We can remove these comments I guess.

packages/ui/.babelrc.json
```diff
@@ -0,0 +1,9 @@
+{
+    "sourceType": "unambiguous",
+    "presets": [
+        ["@babel/preset-env"],
```
guerrap, line 4: We can avoid the `[]` since we are not passing any configuration object to the `preset-env`.

packages/ui/src/components/data-display/chip/index.stories.tsx
```diff
@@ -1,21 +1,20 @@
 import React from "react";
-import { ComponentMeta, Story } from "@storybook/react";
+import type { Meta, StoryObj } from "@storybook/react";
```
guerrap, line 2: Is the `import type` needed here?

packages/frontend/src/constants/index.ts
luzzif: Addressed in 287eedfe63c6938294fe51b84384501ca05a2624

packages/ui/.babelrc.json
luzzif: Addressed in f087b5c11cffb39da92e403c638b77059e5d26bc

packages/ui/src/components/data-display/chip/index.stories.tsx
luzzif: Yeah. Actually, we should start using `import type` always when importing types. The actual import will be erased at runtime and only used at compile time, resulting in a potential reduction of bundle size.

packages/frontend/src/components/ui/kpi-token-card/index.tsx
luzzif: Addressed in 754196f261b2413e7c77ef23f41e7146cf1fca55

guerrap (approved):

PR #251, opened 2023-05-04
Move all fathom to frontend package

guerrap:


packages/frontend/.env.example
```diff
@@ -1,3 +1,6 @@
 REACT_APP_INFURA_PROJECT_ID=
 REACT_APP_WALLETCONNECT_PROJECT_ID=
-REACT_APP_FATHOM_SITE_ID=
\ No newline at end of file
+REACT_APP_FATHOM_SITE_ID=
+
+export REACT_APP_FATHOM_SITE_ID=""
```
luzzif, line 5: Not sure export is needed here. The frontend picks up envs with dotenv (have a look at the `check-env` script)

packages/frontend/CONTRIBUTING.md
```diff
@@ -43,6 +43,16 @@ After you have these envs, you should create a `.env` file in the root of the
 package (or copy the `.env.example` available at the root of the package) and
 paste the values there.
 
+The `.env` file should also `export` 2 additional envs that will be used by the
```
luzzif, line 46: We need to change this text once the exports above are removed

packages/frontend/scripts/sync-fathom-events/index.ts
```diff
@@ -44,7 +44,7 @@ const syncFathomEvents = async (siteId?: string, authToken?: string) => {
 
         const syncedEvents = [...fathomSiteEvents, ...newSiteEvents];
         await writeFile(
-            `${__dirname}/../../src/analytics/generated.ts`,
```
luzzif, line 47: WHat about having a `generated` folder with a `fathom` module in it instead? The idea being that if in the future we need to have more autogenerated files we can just put them there. It also makes it easier to ignore generated code as we can ignore the entire folder instead of each single file (and as such we should update gitignore, prettierignore and the ESLint config)

luzzif (changes requested):

luzzif (approved):

luzzif (approved):

PR #253, opened 2023-05-05
feat: arbitrum goerli support and connectors simplification

luzzif:


packages/frontend/src/connectors/frame/index.ts
```diff
@@ -100,15 +71,15 @@ export class FrameConnector extends Connector<JsonRpcProvider, unknown> {
         return chainId;
     }
 
-    public async getProvider({
-        chainId,
-        create,
-    }: { chainId?: number; create?: boolean } = {}) {
-        if (!this.provider || chainId || create) {
+    public async getProvider(config?: { chainId?: number }) {
+        if (!this.provider) {
```
guerrap, line 75: Should we add an `if(!chiandId) return` before this? Or it's not required?

luzzif: Nono by default it will connect to the currently selected chain in Frame. It's the intended behavior

guerrap (approved):

PR #254, opened 2023-05-10
Remove wagmi peer dependency from UI package

guerrap:
Refactor the `erc20-token-picker` to work properly without the need for `wagmi`.


packages/ui/src/components/evm/erc20-token-picker/hooks/useSearchedTokens.ts
```diff
@@ -1,63 +1,24 @@
...
 import { TokenInfoWithBalance, TokenListWithBalance } from "../types";
 
 export const useSearchedTokens = (
     debouncedQuery?: string,
     chainId?: number,
-    selectedList?: TokenListWithBalance,
-    withBalances?: boolean,
-    connectedAccountAddress?: string
-): { tokens: TokenInfoWithBalance[]; loadingBalances: boolean } => {
-    const cachedImportedTokens = cachedTokenInfoWithBalanceInChain(chainId);
```
luzzif, line 18: Why did we remove so much stuff from here? The cached tokens were useful, otherwise the tokens imported by the user won't be persisted, and will be removed each time the user closes the modal.

packages/ui/src/components/evm/erc20-token-picker/index.tsx
```diff
@@ -32,12 +25,11 @@ export interface ERC20TokenPickerProps {
     onDismiss?: () => void;
     loading?: boolean;
     selectedToken?: TokenInfoWithBalance | null;
+    searchQueryObserver?: (query: string) => void;
```
luzzif, line 28: Isn't it more consistent to rename this to `onSearchQueryChange`?

packages/ui/src/components/evm/erc20-token-picker/hooks/useSearchedTokens.ts
guerrap: The caching functionality has been moved to the erc20 token template, where the `erc20-token-picker` is used. I removed it because the cached token logic, when caching an imported token, relied on `wagmi` (the `useImportableToken` that fetched the token info).

Now the cache it's implemented outside of the component. If we feel we're removing too much from the UI component, I can try to think of another way to tackle this.

packages/ui/src/components/evm/erc20-token-picker/index.tsx
guerrap: Fixed in ec5dd5c9a16aab89aac8b5497ad488db8a3ab2ad.

packages/ui/src/components/evm/erc20-token-picker/hooks/useSearchedTokens.ts
luzzif: Let's just leave it like this for now I'd say

luzzif (approved):

PR #255, opened 2023-05-11
feat(react): typing improvements and pass resolved template to template component

luzzif:


packages/react/src/hooks/useTemplateModule.ts
```diff
@@ -1,25 +1,38 @@
...
+    T extends TemplateType
+>(
+    entity: E,
+    type: T,
     template?: ResolvedTemplate,
     entry?: string
-) => {
+): {
+    loading: boolean;
+    Component: ComponentType<E, T> | null;
```
guerrap, line 31: Why `Component` with the capital `C`?

guerrap (approved):

PR #257, opened 2023-05-12
feat(frontend): improve env settings and build conditions

luzzif:
- define different build command for each build mode (production, staging and library)
- update turbo.json config to reflect new commands
- rename check-env script to config-react-env and avoid requiring certain envs in library and staging mode (fathom tracking and walletconnect support becomes optional in staging mode and not present in library mode)
- split out fathom initialization in its own component that is only loaded in production mode
- enable service worker in staging mode


packages/frontend/CONTRIBUTING.md
```diff
@@ -27,31 +27,34 @@ yarn install
...
+  production bundle). You can get one by registering
   [here](https://app.usefathom.com).
 
 After you have these envs, you should create a `.env` file in the root of the
 package (or copy the `.env.example` available at the root of the package) and
 paste the values there.
 
-The `.env` file should also define 2 additional envs that will be used by the
-`build` command to generate code related to Fathom.
+The `.env` file must also define 2 additional envs that will be used by the
```
guerrap, line 49: We can specify that the `FATHOM` envs are not required if fathom won't be enabled.

luzzif: Yeah but the problem is they are required, otherwise the Fathom code won't be generated and even though it is not used the app will look for it, breaking.

packages/frontend/scripts/config-react-env.js
```diff
@@ -0,0 +1,40 @@
...
+    );
+    process.exit(1);
+}
+
+const requireEnv = (envName) => {
+    const env = process.env[envName];
+    if (!env) throw new Error(`a ${envName} env is required`);
+};
+
+const isEnvDefined = (envName) => !!process.env[envName];
```
guerrap, line 26: I think we can avoid defining this function since it's only used once.

packages/frontend/CONTRIBUTING.md
guerrap: Ok, I'll update this section once the lib is updated, to actually make the envs optional, by generating "mocked" code.

guerrap (dismissed):

packages/frontend/scripts/config-react-env.js
luzzif: Addressed in 9bf473c9156b237eecc3cb427014faf2ce96eab4

packages/frontend/scripts/sync-fathom-events/index.ts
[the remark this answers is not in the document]
luzzif: Addressed in 5c85a9cfa59f6f1c0d5d7684308f3b53488b29dd

guerrap (approved):

PR #264, opened 2023-05-17
Add new typography definitions

guerrap:
Once this PR is merged and released I'll open 2 more PRs for the template repositories, to use the new typography variants.


packages/frontend/src/pages/home/latest-campaigns-section/index.tsx
```diff
@@ -21,7 +21,7 @@ export const LatestCampaignsSection = () => {
         <div className="relative flex flex-col gap-16">
             {/* TODO: add i18n */}
```
luzzif, line 22: Is this TODO still applicable? Maybe we should move it below where there's a need for `i18n` (the "view all campaigns text").

packages/ui/src/components/input/commons/index.tsx
```diff
@@ -194,7 +193,7 @@ export const BaseInputWrapper = ({
     className,
     children,
 }: BaseInputWrapperProps): ReactElement => {
-    const [infoIcon, setInfoIcon] = useState<SVGSVGElement | null>(null);
+    const infoIconRef = useRef<HTMLDivElement | null>(null);
```
luzzif, line 196: We can't do this otherwise the popover won't update position if the anchor element is moved (yeah I know that kinda doesn't happen anyway, but [here](https://popper.js.org/react-popper/v2/#example)'s what the official popper doc has to say)

luzzif (changes requested):

packages/frontend/src/pages/home/latest-campaigns-section/index.tsx
guerrap: Oh ye, I guess we can directly solve it, I'll add the translation.

guerrap: Fixed in bd09652fd0d9567326d5ff6221b1c2ef5c9187b2.

packages/ui/src/components/input/commons/index.tsx
guerrap: Fixed in 89da18cc23654778f43a45577bf2968c8f3c57c0.

luzzif: Why is the `as` needed? We need to change the type of the `useState` to `HTMLDivElement` from `SVGSVGElement`.

guerrap: Oh you're right, my bad; fixed in af37f2978c4b83b36c4d1c6313db893c8f4f91ca.

luzzif (approved):

PR #273, opened 2023-05-30
fix(ui): minor style fixes

luzzif:


packages/ui/src/components/input/date-time/picker/index.tsx
```diff
@@ -12,7 +12,8 @@ const cellListStyles = mergedCva(
         "cui-w-full md:cui-w-10",
         "cui-py-2",
         "cui-overflow-y-auto",
-        "cui-scrollbar",
+        "scrollbar-none",
```
guerrap, line 15: Are we missing a `cui-` here?

guerrap (approved):

PR #282, opened 2023-06-13
Fetch latest KPI tokens

guerrap:


packages/frontend/src/hooks/useLatestKPITokens.ts
```diff
@@ -25,20 +25,21 @@ export function useLatestKPITokens(limit = 5): {
...
+                    limit,
+                });
             const kpiTokens = await Fetcher.fetchKPITokens({
                 publicClient,
                 preferDecentralization,
                 addresses: kpiTokenAddresses,
             });
 
-            return Object.values(kpiTokens).reverse();
+            return preferDecentralization
```
luzzif, line 40: Not much of a fan of this... I'd prefer if both the onchain and subgraph versions would return arrays ordered the same way. It's probably ok to return them ordered by creation time in asc order and then do a reverse here.

luzzif (changes requested):

packages/sdk/src/fetcher/on-chain/index.ts
```diff
@@ -63,6 +64,26 @@ class Fetcher implements IPartialCarrotFetcher {
...
+        const toIndex = await this.fetchKPITokensAmount({ publicClient });
+        const fromIndex = toIndex - finalLimit > 0 ? toIndex - finalLimit : 0;
+        return (
+            await publicClient.readContract({
+                abi: FACTORY_ABI,
+                address: chainAddresses.factory,
+                functionName: "enumerate",
+                args: [BigInt(fromIndex), BigInt(toIndex)],
+            })
+        ).slice();
```
luzzif, line 84: I'm wondering if instead of slicing (I'm guessing this is to make the end result not readonly) we should just do an explicit casting. Because there's nothing in the code making this readonly at runtime. It's just a Typescript compile time thing. So an `as` should do the trick.

luzzif (changes requested):

luzzif (approved):

PR #285, opened 2023-06-15
fix(ui): allow non serializable select values

luzzif:


packages/ui/src/components/input/select/index.tsx
```diff
@@ -121,15 +121,10 @@ function Component<T extends SelectOption<ValueType>>(
...
-    const handlePick = useCallback(
-        (event: SpecificMouseEvent<HTMLLIElement>) => {
-            if (!event.target) return;
-            const value = (event.target as HTMLLIElement).dataset.value;
-            if (!!onChange && !!value) onChange(JSON.parse(value) as T);
-            setOpen(false);
-        },
-        [onChange]
-    );
+    const getPickHandler = (option: T) => () => {
```
guerrap, line 124: Wondering if we want to keep the `useCallback`.

luzzif: Addressed in dd3e6094602f440f4ca5080639c5e074c342ad6d

packages/ui/src/components/input/select/index.tsx
```diff
@@ -121,10 +120,13 @@ function Component<T extends SelectOption<ValueType>>(
         setOpen(!open);
     }, [open]);
 
-    const getPickHandler = (option: T) => () => {
-        if (onChange) onChange(option);
-        setOpen(false);
-    };
+    const getPickHandler = useCallback(
+        (option: T) => () => {
+            if (onChange) onChange(option);
```
guerrap, line 125: Do we need to check if the `onChain` function exists? It's a required prop so maybe it can be removed.

luzzif: You're right, addressed in 4f81b738811921bd08c9548bbeb3f139757e949d

guerrap (approved):

PR #299, opened 2023-07-13
feat: introduce new ipfs pinning proxy

luzzif:


packages/frontend/src/components/authenticate/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const [loading, setLoading] = useState(false);
+
+    const { data: signedLoginMessage, signMessage } = useSignMessage();
+
+    useEffect(() => {
+        const fetchData = async () => {
+            if (!address || !signedLoginMessage) return;
+            setLoading(true);
+            try {
+                if (!address || !signedLoginMessage) return;
```
guerrap, line 26: We can remove this `if` since we're checking for the same condition above.

guerrap (changes requested):

packages/frontend/src/components/authenticate/index.tsx
luzzif: Fixed in 00b83a648f5fa57aab1827856ac5814678ba8fda

PR #300, opened 2023-07-14
Avoid resolving all the KPI tokens at once

guerrap:


packages/frontend/src/pages/campaigns/index.tsx
```diff
@@ -65,13 +49,10 @@ export const Campaigns = () => {
     useEffect(() => {
         if (loading || kpiTokens.length > 0) return;
         setKpiTokens(Object.values(response));
-        setKpiTokensReady(true);
```
guerrap: I think we should keep `kpiTokensReady`, otherwise the `Empty` component would be rendered for a second, between the time the KPI tokens are fetched (`loading: false`) and the `setKpiTokens` actually completes.

MilanVojnovic95:
Tested and looks ok to me 👍

luzzif (approved):

PR #326, opened 2023-08-02
Replace __DEV__ env with __PROD__

guerrap:


packages/frontend/rollup.config.mjs
```diff
@@ -26,7 +26,7 @@ export default [
                 values: {
                     __LIBRARY_MODE__: JSON.stringify(true),
                     __STAGING_MODE__: JSON.stringify(false),
-                    __DEV__: JSON.stringify(false),
+                    __PROD__: JSON.stringify(true),
```
luzzif, line 29: Pretty sure this is wrong. Dev was set to false here because staging mode is different from dev  mode, but it's not prod either. this needs to be set to false

guerrap: Fixed in 25cafe90021b1fdf6f6e04da28db303445f2468c.

luzzif (approved):

PR #334, opened 2023-08-11
chore(react): flatten file structure and make library tree shakeable

luzzif:


packages/react/rollup.config.mjs
```diff
@@ -15,10 +56,11 @@ export default [
             typescript(),
         ],
         output: {
-            file: resolve("dist/index.js"),
+            dir: resolve("./dist"),
+            preserveModules: true,
+            preserveModulesRoot: "src",
             format: "es",
             sourcemap: true,
-            inlineDynamicImports: true,
```
guerrap, line 21: Is this no longer needed?

luzzif: Seems like it isn't, yes

PR #363, opened 2023-09-02
chore: integrate scroll sepolia and carrot-kpi contracts lib

luzzif:
And upgrade dependencies


packages/frontend/src/standalone-setup/index.ts
```diff
@@ -21,9 +21,9 @@ const RPC_BY_CHAIN: Record<ChainId, RPCConfig> = {
         http: "https://rpc.ankr.com/gnosis",
         webSocket: "wss://rpc.gnosischain.com/wss",
     },
-    [ChainId.SCROLL_TESTNET]: {
-        http: "https://alpha-rpc.scroll.io/l2",
-        webSocket: "wss://alpha-rpc.scroll.io/l2/ws",
+    [ChainId.SCROLL_SEPOLIA]: {
+        http: "https://sepolia-rpc.scroll.io/",
```
guerrap, line 25: Let's remove the ending `/` here, so it's the same as before.

guerrap (approved):

packages/frontend/src/standalone-setup/index.ts
luzzif: Addressed in b0ca6c62ee0e09dc7d2d949b039bbbc64eaebd82

guerrap (dismissed):

PR #364, opened 2023-09-04
Add prop to make select searchable

guerrap:


packages/ui/src/components/select.tsx
```diff
@@ -81,9 +87,11 @@ export type SelectProps<T extends SelectOption<ValueType>> = {
     id?: string;
     options: T[];
     value: T | null;
+    search?: boolean;
     onChange: (value: T) => void;
     renderOption?: (value: T) => ReactElement;
     loading?: boolean;
+    noResultsText?: string;
```
luzzif, line 94: To keep consistency with other components needing messages, I'd use a `messages` object prop with key `noText`, wdyt?

luzzif (changes requested):

packages/ui/src/components/select.tsx
guerrap: Yes, definitely.

luzzif (approved):

PR #373, opened 2023-09-18
feat(ui): use virtualized list for select component

guerrap:


packages/ui/src/components/select.tsx
```diff
@@ -9,38 +9,52 @@ import React, {
...
+        "cui-rounded-xxl",
+        "cui-border",
+        "cui-border-black",
+        "cui-max-h-48",
+        "cui-bg-white",
+        "dark:cui-bg-black",
+        "dark:cui-border-white",
+        "cui-z-20",
+        "cui-overflow-hidden",
+        "cui-h-full",
```
luzzif, line 33: Why `cui-h-full` here and then `cui-h-fit` when empty? What happens if the select has less than 4 options? Will the dropdown resize accordingly?

packages/ui/src/components/select.tsx
```diff
@@ -9,38 +9,52 @@ import React, {
...
-    "cui-overflow-y-auto",
+const emptyOptionListStyles = mergedCva([
+    "cui-flex",
+    "cui-justify-center",
+    "cui-items-center",
+    "cui-p-3",
+    "cui-h-12",
 ]);
 
+const optionListStyles = mergedCva(["cui-scrollbar"]);
```
luzzif, line 52: I honestly think we should probably style the scrollbar so that it's a bit less big, even though that's probably what cui-scrollbar already does?

packages/ui/src/components/select.tsx
```diff
@@ -9,38 +9,52 @@ import React, {
...
+
 const optionStyles = mergedCva(
     [
         "cui-cursor-pointer",
         "cui-p-3",
-        "cui-h-12",
```
luzzif, line 43: Why remove the `h-12`?

packages/ui/src/components/select.tsx
```diff
@@ -212,54 +213,76 @@ function Component<T extends SelectOption<ValueType>>(
...
+                    <AutoSizer disableWidth>
+                        {({ height }) => {
                             return (
-                                <li
-                                    className={optionStyles({
-                                        picked: value?.value === option.value,
-                                        className: className?.option,
+                                <FixedSizeList
+                                    height={height}
+                                    width={"auto"}
```
luzzif, line 239: Does it make sense to use width auto here and then later set it through inline styles?

luzzif (changes requested):

packages/ui/src/components/select.tsx
guerrap: Yes, `cui-scrollbar` gets the style from the `global.css` in the ui.

packages/ui/src/components/select.tsx
guerrap: Fixed in 3ed97c23e9d9e240a6ab53d46e81013989c7c9f6.

packages/ui/src/components/select.tsx
guerrap: Fixed in 3ed97c23e9d9e240a6ab53d46e81013989c7c9f6.

packages/ui/src/components/select.tsx
guerrap: Fixed in 3ed97c23e9d9e240a6ab53d46e81013989c7c9f6.

PR #374, opened 2023-09-18
feat(ui): more flexible feedback box and selectable chips with size variants 

luzzif:


packages/ui/src/components/feedback-box.tsx
```diff
@@ -0,0 +1,75 @@
...
+}: FeedbackBoxProps) => {
+    return (
+        <div
+            className={rootStyles({
+                variant,
+                border,
+                className: className?.root,
+            })}
+        >
+            <div className={titleContainerStyles()}>
```
guerrap, line 60: We could add a `titleContainer` class name here, to make this customizable.

guerrap (approved):

PR #395, opened 2023-09-29
feat: handle different staging/prod clusters/services

luzzif:


packages/shared-state/src/reducers/preferences/reducer/index.ts
```diff
@@ -14,7 +14,7 @@ const initialState: PreferencesState = {
     // TODO: use system as a default once the dark theme is available
     theme: "light",
     preferDecentralization: false,
-    ipfsGatewayURL: "https://gateway.carrot-kpi.dev",
+    ipfsGatewayURL: "https://gateway.api.carrot.community",
```
guerrap, line 17: I think we should use the new `getServiceURL` here, no?

packages/frontend/src/constants/index.ts
```diff
@@ -25,10 +32,8 @@ export const SUPPORTED_CHAINS: Record<ChainId, AugmentedChain> = {
         ...gnosis,
         logo: GnosisLogo,
         iconBackgroundColor: "#04795b",
-        enabled: true,
+        enabled: !__STAGING_MODE__,
```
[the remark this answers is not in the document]
luzzif: You're right, addressed in 97e540018ce5d7b369f711cf9e7a4ce4c055c0f4. I've also disabled testnets in prod

packages/shared-state/src/reducers/preferences/reducer/index.ts
luzzif: The thing is that this addess will be overridden anyway on the app's mount. It's a bit of a weird setup

guerrap (approved):

PR #398, opened 2023-10-04
fix(sdk): drastically reduce number of eth_getChainId calls

luzzif:


packages/frontend/src/standalone-entrypoint.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { HashRouter } from "react-router-dom";
+import { HostStateProvider } from "./state/connector";
+import { ReactSharedStateProvider } from "@carrot-kpi/shared-state";
+import {
+    useSetDevMode,
+    useSetIPFSGatewayURL,
+    useSetStagingMode,
+} from "@carrot-kpi/react";
+
+console.log(
```
guerrap, line 38: Is this log needed?

guerrap (approved):

PR #414, opened 2023-10-06
feat(subgraph): add new sepolia subgraph

luzzif:


packages/subgraph/networks.json
```diff
@@ -12,5 +12,33 @@
             "address": "0xc9E426B468d334AF6208ef9b3bA5f599d1417c6e",
             "startBlock": 28680520
         }
+    },
+    "scroll-sepolia": {
```
guerrap, line 16: Scroll Sepolia has a subgraph?

packages/sdk/src/commons.ts
```diff
@@ -38,8 +38,9 @@ export const CHAIN_ADDRESSES: Record<ChainId, ChainAddresses> = {
 
 export const SUBGRAPH_URL: Record<ChainId, string | null> = {
     [ChainId.GNOSIS]:
-        "https://api.thegraph.com/subgraphs/name/carrot-kpi/carrot-kpi-gnosis",
-    [ChainId.SEPOLIA]: null,
+        "https://api.thegraph.com/subgraphs/name/carrot-kpi/carrot-gnosis",
+    [ChainId.SEPOLIA]:
+        "https://api.thegraph.com/subgraphs/name/carrot-kpi/carrot-sepolia",
     [ChainId.SCROLL_SEPOLIA]: null,
```
guerrap, line 44: If Scroll Sepolia has a subgraph aren't we missing the url here?

luzzif: It doesn't have a Carrot subgraph deployment. TheGraph supports Scroll Sepolia but I thought it would make sense to have at least one network without a subgraph for testing purposes (for example something that we're currently missing is the removal of the decentralization mode toggle for networks that don't have a Carrot subgraph deployment)

packages/subgraph/networks.json
luzzif: Not really but the new autogen script that generates this file doesn't make distinctions. If a network has a Carrot deployment it ends up here, even if TheGraph doesn't support the network. I think it's fine as the ultimate source of truth on where subgraphs are deployed are the package.json scripts

guerrap (approved):

PR #431, opened 2023-10-12
Campaigns page

### LINKED ISSUE #422: All campaigns page/ Filleting is not working properly
As you can see here on this photo :

The Newest and Finalized options are selected, 
however, one campaign is left out, which has the 
same finalization time and is also finalized.

Screenshot 2023-10-10 at 5 00 40 PM

A finalized campaign that was left out.
Screenshot 2023-10-10 at 5 16 59 PM

guerrap:
Fixes #422.

### DIFF
--- a/packages/frontend/src/components/ui/navbar/index.tsx
+++ b/packages/frontend/src/components/ui/navbar/index.tsx
@@ -58,7 +58,7 @@ export const Navbar = ({
                 links={links}
                 onNavbarClose={() => setOpen(false)}
             />
-            <div className="flex justify-center px-4 xl:px-32 bg-orange">
+            <div className="flex justify-center px-4 xl:px-32 bg-transparent">
                 <div className="w-full max-w-screen-2xl relative flex items-center justify-between py-5 md:py-8 xl:py-11">
                     <div className="flex gap-4 items-center">
                         <div className="flex items-center">
--- a/packages/frontend/src/components/ui/navbar/index.tsx
+++ b/packages/frontend/src/components/ui/navbar/index.tsx
@@ -58,7 +58,7 @@ export const Navbar = ({
                 links={links}
                 onNavbarClose={() => setOpen(false)}
             />
-            <div className="flex justify-center px-4 xl:px-32 bg-transparent">
+            <div className="flex justify-center px-4 xl:px-32 bg-orange">
                 <div className="w-full max-w-screen-2xl relative flex items-center justify-between py-5 md:py-8 xl:py-11">
                     <div className="flex gap-4 items-center">
                         <div className="flex items-center">
--- a/packages/react/src/hooks/useResolvedKPITokens.ts
+++ b/packages/react/src/hooks/useResolvedKPITokens.ts
@@ -57,10 +57,10 @@ export function useResolvedKPITokens(params?: ResolvedKPITokensParams): {
                     }),
                 );
             } catch (error) {
-                console.error("error fetching kpi tokens", error);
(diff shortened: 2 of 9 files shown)


packages/frontend/src/components/ui/navbar/index.tsx
```diff
@@ -58,7 +58,7 @@ export const Navbar = ({
                 links={links}
                 onNavbarClose={() => setOpen(false)}
             />
-            <div className="flex justify-center px-4 xl:px-32 bg-transparent">
+            <div className="flex justify-center px-4 xl:px-32 bg-orange">
```
luzzif, line 61: I'm not sure we can do this as the creation form has a green bg for example and the navbar has to match. Previously there was a prop to the navbar that dictated the bg color which I removed. At this point I'd restore it.

packages/react/src/hooks/useResolvedKPITokens.ts
```diff
@@ -0,0 +1,77 @@
...
+    >([]);
+
+    const [loading, setLoading] = useState(true);
+
+    useEffect(() => {
+        let cancelled = false;
+        async function fetchData(): Promise<void> {
+            if (!cancelled) setLoading(true);
+
+            try {
```
luzzif, line 29: Let's use a `finally` here to set loading to false, it's just safer and gets always executed if anything happens.

packages/react/src/hooks/useResolvedKPITokens.ts
```diff
@@ -0,0 +1,77 @@
...
+                        try {
+                            const resolved = (
+                                await Fetcher.resolveKPITokens({
+                                    ipfsGatewayURL,
+                                    kpiTokens: [kpiToken],
+                                })
+                            )[kpiToken.address];
+                            if (!resolved) return;
+                            if (!cancelled)
+                                setResolvedKPITokens((previousState) => [
```
luzzif, line 47: I'm wondering if any time based sorting to display KPI tokens in the campaigns list is perfirmed on the campaigns page or not, because this can result in potentially out of order resolutions.

packages/react/src/hooks/useResolvedKPITokens.ts
```diff
@@ -0,0 +1,77 @@
...
+                        } catch (error) {
+                            console.error(
+                                `error resolving kpi token at address ${kpiToken.address}`,
+                                error,
+                            );
+                        }
+                    }),
+                );
+            } catch (error) {
+                console.error("error fetching kpi tokens", error);
```
luzzif, line 60: Let's maybe have "error fetching resolved kpi tokens" here

luzzif (changes requested):

packages/frontend/src/components/ui/navbar/index.tsx
guerrap: Fixed in 0a2477eb92e01e1cc5dc6c1be6be4d3d46d1f46c.

packages/react/src/hooks/useResolvedKPITokens.ts
guerrap: Fixed in ba94fdd708c0c95ea31e879ddda10c31848ce4d8.

packages/react/src/hooks/useResolvedKPITokens.ts
guerrap: Fixed in ba94fdd708c0c95ea31e879ddda10c31848ce4d8.

PR #434, opened 2023-10-13
feat(ui): add error and different info variants to feedback box

guerrap:


packages/ui/src/components/feedback-box.tsx
```diff
@@ -6,8 +6,10 @@ import Warning from "../icons/warning";
 const rootStyles = mergedCva(["cui-p-4", "cui-border", "cui-border-black"], {
     variants: {
         variant: {
-            info: ["cui-bg-green"],
+            infoGreen: ["cui-bg-green"],
```
luzzif, line 9: Why not `success` for green and `info` for blue?

luzzif (changes requested):

packages/ui/src/components/feedback-box.tsx
guerrap: Fixed in a1c6692b33a9b98bb57fabedb1a7356856f47696.

luzzif (approved):

PR #451, opened 2023-10-23
fix(frontend): fix creation allowance check for connected addresses

luzzif:


packages/frontend/src/hooks/useIsCreatorAllowed.ts
```diff
@@ -0,0 +1,30 @@
...
+                functionName: "permissionless",
+            },
+            {
+                address: factoryAddress,
+                abi: FACTORY_ABI,
+                functionName: "creatorAllowed",
+                args: [creator as Address],
+            },
+        ],
+        enabled: chainId in ChainId,
```
guerrap, line 23: Should we add `creator` in the condition here?

luzzif: Addressed in a957f5af4b51ebbb78297cca3d341884afcefd2e

guerrap (approved):

PR #472, opened 2023-10-30
feat: first base implementation of a drafts feature

luzzif:


packages/react/src/types/templates.ts
```diff
@@ -46,37 +54,50 @@ export interface OracleInitializationBundle {
...
+) => void;
+
+export type OracleCreationFormProps<S extends Serializable = Serializable> =
+    TemplateComponentProps & AdditionalRemoteOracleCreationFormProps<S>;
+
+export interface AdditionalRemoteKPITokenCreationFormProps<
+    S extends Serializable = Serializable,
+> {
+    state: S;
+    onChange: KPITokenChangeCallback<S>;
```
guerrap, line 88: Should we rename this to be more specific? To what `onChange` event this function is linked to in the `KPITokenCreationForm`?

luzzif: The internal state's change. I think it's probably clear enough when looking at the `KPITokenChangeCallback` type. It's the same name and functionality in the oracle template type too I think.

guerrap (approved):

PR #487, opened 2023-11-02
fix(ui): fix validation of min and max date in date time picker

luzzif:


packages/ui/src/components/date-time-input/picker/index.tsx
```diff
@@ -93,24 +93,20 @@ export const DateTimePicker = ({
...
-                if (dayjs(newMin).isAfter(dayjs(newMax), "seconds")) {
-                    newMin = newMax;
-                    console.warn("inconsistent min and max values", {
-                        min: newMin?.toISOString(),
-                        max: newMax?.toISOString(),
-                    });
-                }
-                return newMin;
-            });
+        setMin(updatedMin);
```
guerrap, line 99: Wondering if these `setState`s, can trigger an infinite loop since they update `min/max` values, that are also in the dependency array of the `useEffect`.

luzzif: Yes but the first time they're triggered when a min or max date changes, and those are set in the state, triggering the second run. In the second run the state and passed min and max dates are the same so nothing happens and no useEffect is triggered anymore

guerrap (approved):

PR #504, opened 2023-11-07
feat: add polygon mumbai support

luzzif:


packages/subgraph/package.json
```diff
@@ -1,27 +1,30 @@
...
+        "@carrot-kpi/contracts": "^0.9.0",
+        "@graphprotocol/graph-cli": "0.60.0",
         "@graphprotocol/graph-ts": "0.31.0",
         "eslint-config-custom": "*",
-        "matchstick-as": "0.5.2",
+        "matchstick-as": "0.6.0",
         "rimraf": "^5.0.5",
-        "ts-node": "^10.9.1"
+        "ts-node": "^10.9.1",
+        "tsconfig": "*"
```
guerrap, line 28: Is this desired? To have any version?

packages/subgraph/subgraph.yaml
```diff
@@ -7,11 +7,11 @@ features:
 dataSources:
   - kind: ethereum/contract
     name: Factory
-    network: gnosis
+    network: mumbai
```
guerrap, line 10: Why did gnosis get replaced here?

packages/ui/src/utils/components.tsx
```diff
@@ -12,8 +12,9 @@ if (theme) {
 
 const twMergeWithConfig = extendTailwindMerge({
     prefix: "cui-",
-    theme,
-    classGroups,
+    extend: {
```
guerrap, line 15: Is theme not needed anymore?

packages/subgraph/package.json
luzzif: `tsconfig` is a local dependency internal to the monorepo, so with `yarn` that's how you specify it.

packages/subgraph/subgraph.yaml
luzzif: That's a side effect of how the TheGraph's library handles the `--network` parameter passed when deploying a subgraph. It basically replaces the `network` and `address` fields in the `subgraph.yaml` file in place. The end result is that each time you do a deployment on a different network this file will be updated. So it's expected.

packages/ui/src/utils/components.tsx
luzzif: I tried to start up the frontend locally and didn't notice any difference really. I'm also not noticing any difference in the UI preview above.

guerrap (approved):

PR #509, opened 2023-11-08
feat(qa): adding selectors for smoke test

MilanVojnovic95:
- adding selectors for smoke test checks

### DIFF
--- a/packages/frontend/src/components/permissioned.tsx
+++ b/packages/frontend/src/components/permissioned.tsx
@@ -20,12 +20,14 @@ export const Permissioned = ({ onBack }: PermissionedProps) => {
                     <>
                         <ErrorIcon className="w-52 h-52" />
                         <Typography
+                            data-testid="permission-mode-enabled-text"
                             variant="h4"
                             className={{ root: "text-center" }}
                         >
                             {t("permissioned.mode.title")}
                         </Typography>
                         <Typography
+                            data-testid="permission-mode-description-text"
                             className={{
                                 root: "text-center mb-4",
                             }}
@@ -34,6 +36,7 @@ export const Permissioned = ({ onBack }: PermissionedProps) => {
                         </Typography>
                         <div className="flex gap-4">
                             <Button
+                                data-testid="back-button"
                                 size="small"
                                 variant="secondary"
                                 onClick={onBack}
@@ -41,6 +44,7 @@ export const Permissioned = ({ onBack }: PermissionedProps) => {
                                 {t("permissioned.mode.back")}
                             </Button>
                             <Button
+                                data-testid="join-discord-button"
                                 size="small"
                                 href={DISCORD_LINK}
                                 target="_blank"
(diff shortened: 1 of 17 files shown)


packages/frontend/src/components/permissioned.tsx
```diff
@@ -20,12 +20,14 @@ export const Permissioned = ({ onBack }: PermissionedProps) => {
                     <>
                         <ErrorIcon className="w-52 h-52" />
                         <Typography
+                            data-testid="permission-mode-enabled-text"
```
guerrap, line 23: Should we rename this to `permission-mode-enabled-title-text`, to be more specific?

packages/frontend/src/components/permissioned.tsx
```diff
@@ -34,13 +36,15 @@ export const Permissioned = ({ onBack }: PermissionedProps) => {
                         </Typography>
                         <div className="flex gap-4">
                             <Button
+                                data-testid="back-button"
```
guerrap, line 39: Same as above, we could rename this to  `permission-mode-back-button`.

packages/frontend/src/components/ui/empty.tsx
```diff
@@ -57,10 +57,16 @@ export const Empty = ({
...
-                <Typography variant={titleVariant || "h3"}>
+                <Typography
+                    data-testid="empty-title-text"
+                    variant={titleVariant || "h3"}
+                >
                     {t("empty.title")}
                 </Typography>
-                <Typography variant={descriptionVariant || "base"}>
+                <Typography
+                    data-testid="no-data-found-text"
```
guerrap, line 67: Same here `empty-no-data-found-text`.

packages/frontend/src/components/ui/footer.tsx
```diff
@@ -5,13 +5,18 @@ import { Typography } from "@carrot-kpi/ui";
 
 const FooterLinks = ({ title, links }: FooterLink) => (
     <ul className="w-40">
-        <Typography className={{ root: "text-white" }} uppercase>
+        <Typography
+            data-testid={`${title}-text`}
```
guerrap, line 9: Same here, I think as a rule we could prefix the test ids with the component name, here it would be `footer-...`.

packages/frontend/src/components/permissioned.tsx
MilanVojnovic95: agreed

packages/frontend/src/components/ui/empty.tsx
MilanVojnovic95: agreed

packages/frontend/src/components/permissioned.tsx
MilanVojnovic95: agreed

packages/frontend/src/components/ui/footer.tsx
MilanVojnovic95: agreed

guerrap (approved):

PR #511, opened 2023-11-14
feat: initial smoke test

MilanVojnovic95:
- added smoke test for carrot home page
- checking text 
- checking redirections

Note: currently will work on local since I have added additional selector

It can be optimizes and improved but I wanted to have some checks for start and improve it later


packages/frontend/e2e/tests/connect-wallet.test.ts
guerrap, line 1: Why is this all commented?

packages/frontend/e2e/utils/data/data.ts
guerrap, line 1: Do we need other files under `/data` other than `data.txs`? If that's not the case I'd remove the folder and just have the `data.ts` file.

packages/frontend/e2e/tests/smoke.test.ts
```diff
@@ -0,0 +1,104 @@
+import { test } from "../utils/fixtures";
+import { networks, footerLinks } from "../utils/data/data";
+/**
```
guerrap, line 3: Let's leave a line between the imports and the rest.

packages/frontend/e2e/utils/fixtures.ts
```diff
@@ -3,10 +3,18 @@ import { initialSetup } from "@synthetixio/synpress/commands/metamask";
 import { setExpectInstance } from "@synthetixio/synpress/commands/playwright";
 import { resetState } from "@synthetixio/synpress/commands/synpress";
 import { prepareMetamask } from "@synthetixio/synpress/helpers";
-
+import { HomePage } from "../utils/pages/homePage";
+/**
```
guerrap, line 7: Same here, new line.

packages/frontend/e2e/utils/pages/homePage.ts
```diff
@@ -0,0 +1,294 @@
+import { BasePage } from "../pages/basePage";
+import {
+    campaignData,
+    urls,
+    networks,
+    templateData,
+    textData,
+    wallets,
+} from "../data/data";
+/**
```
guerrap, line 10: Here too.

packages/frontend/package.json
```diff
@@ -35,12 +35,14 @@
         "config-react-env": "./scripts/config-react-env.js",
         "start:staging": "STAGING=true craco start",
         "start": "craco start",
-        "test:e2e": "yarn prepare-fathom && playwright test"
+        "test:e2e": "yarn prepare-fathom && playwright test",
+        "smoke": "yarn playwright test"
```
guerrap, line 39: Why a new `smoke` script if we run the same `playwright test` like the script above?

packages/frontend/e2e/tests/connect-wallet.test.ts
MilanVojnovic95: I forgot to uncomment it

packages/frontend/e2e/utils/data/data.ts
MilanVojnovic95: I will leave only the `data.ts` file for now

packages/frontend/package.json
MilanVojnovic95: yes I wanted to have separate script just for smoke test and the other one for all tests I'will just edit the existing one to be ` "test:e2e": "yarn prepare-fathom && yarn playwright test"` in order to work

packages/frontend/package.json
```diff
@@ -35,8 +35,7 @@
         "config-react-env": "./scripts/config-react-env.js",
         "start:staging": "STAGING=true craco start",
         "start": "craco start",
-        "test:e2e": "yarn prepare-fathom && playwright test",
```
guerrap: We can omit the `yarn` before `playwright test`. so just have:

`yarn prepare-fathom && playwright test`

MilanVojnovic95: I will edit it now but I am getting error `command not found: playwright test` if I ommit yarn

guerrap: Running it on my machine it works fine without the `yarn`, anyway let's leave it if it causes issues.

guerrap (approved):

PR #536, opened 2023-12-18
KPI token fetch offload

guerrap:


packages/react/src/hooks/useKPIToken.ts
```diff
@@ -22,19 +22,21 @@ export function useKPIToken(params?: KPITokenParams): {
     useEffect(() => {
         let cancelled = false;
         async function fetchData(): Promise<void> {
-            if (!chain || !params?.kpiTokenAddress) return;
+            if (!chain || !params?.kpiTokenAddress || !!kpiToken) return;
```
luzzif, line 25: Let's imagine for a moment that the parameters change but there's a cached `kpiToken` in the state already. What would happen? I'm not 100% sure this scenario can happen in the first place, but maybe we should make this check a bit more robust, like: `kpiToken?.address === params?.address`?

luzzif: Either that, or maybe let's remove the check completely to avoid the potentially infinite loop since `kpiToken` is now both set in this hook and in its dependencies array.

luzzif (changes requested):
Should we completely remove the `...WithData` entities? I'm thinking they are not really used anymore, and single templates will most likely surgically fetch data only and type it separately from the `Oracle` and `KPIToken` entities. Wdyt?

packages/react/src/hooks/useKPIToken.ts
guerrap: Not having that condition would cause an infinite loop since the `kpiToken` gets updated in the `useEffect` (this cause a re-render and then the `useEffect` gets executed again); I think we could improve the check as you suggest.

guerrap: Fixed in 05a8f6f94bc220a79a4167c080b959f95c02b496.

luzzif (approved):

PR #541, opened 2023-12-19
feat(frontend): docker setup to execute e2e tests on ci

luzzif:


docker-compose.e2e.yaml
```diff
@@ -0,0 +1,52 @@
...
+    environment:
+      - RUN_XTERM=no
+      - DISPLAY_WIDTH=1920
+      - DISPLAY_HEIGHT=1080
+    ports:
+      - "8080:8080"
+    networks:
+      - x11
+
+  video:
```
guerrap, line 39: Is this needed in the CI?

guerrap (approved):

docker-compose.e2e.yaml
luzzif: I should have removed this in 153436a53cfa67f5ca3970a5b251c03538e30a94

PR #542, opened 2023-12-19
Add dedicated tsconfig for cypress tests

guerrap:


packages/frontend/rollup.config.mjs
```diff
@@ -35,21 +35,23 @@ export default [
...
                 extract: resolve("./dist/styles.css"),
             }),
             json(),
-            typescript(),
             image(),
         ],
         output: {
-            dir: resolve("dist"),
+            dir: resolve("./dist"),
+            preserveModules: true,
```
luzzif, line 53: Is this actually needed with the new tsconfig?

guerrap: Fixed in 911aae37d4a9927e81cfb9876e0c928171b81f2c.

luzzif (approved):

PR #552, opened 2023-12-22
feat(subgraph): handle new template features registry in kpi token and oracle managers

luzzif:


packages/subgraph/schema.graphql
```diff
@@ -130,6 +148,20 @@ type OracleTemplate @entity {
...
+    allowed: [OracleTemplateFeatureAllowedAccount!]! @derivedFrom(field: "feature")
+}
+
+type OracleTemplateFeatureAllowedAccount @entity {
+    id: Bytes!
+    address: Bytes!
+    feature: OracleTemplateFeature!
+}
+
 type OraclesManager @entity {
```
guerrap, line 165: Aren't we missing the `featuresOwner` and `features` for the `OraclesManager` entity?

packages/subgraph/src/commons.ts
```diff
@@ -60,6 +60,18 @@ export function templateId(onChainId: BigInt, onChainVersion: BigInt): Bytes {
     );
 }
 
+export function featureId(
+    managerAddress: Bytes,
+    templateId: BigInt,
+    featureId: BigInt,
+): Bytes {
+    return managerAddress.concat(
```
guerrap, line 68: Can we add a sort of divider when generating the `featureId`? Something like `manager_address-template_id-feature_id`.

packages/subgraph/src/kpi-tokens-manager.ts
```diff
@@ -76,6 +84,37 @@ export function getTemplate(
...
+        templateId,
+        onChainFeatureId,
+    );
+    let feature = KPITokenTemplateFeature.load(id);
+    if (feature == null) {
+        feature = new KPITokenTemplateFeature(id);
+        feature.featureId = onChainFeatureId;
+        feature.paused = false;
+
+        const oracleManager = getKPITokensManager();
```
guerrap, line 102: `const oracleManager` should be `const kpiTokenManager`, no? Since we are using `getKPITokensManager()`.

packages/subgraph/src/kpi-tokens-manager.ts
```diff
@@ -183,3 +222,92 @@ export function handleUpgradeTemplate(event: UpgradeTemplateEvent): void {
     }
     newTemplate.save();
 }
+
+export function handleSetFeatureSetOwner(event: SetFeatureSetOwnerEvent): void {
+    const oraclesManager = getKPITokensManager();
```
guerrap, line 227: Same as above.

packages/subgraph/src/kpi-tokens-manager.ts
```diff
@@ -183,3 +222,92 @@ export function handleUpgradeTemplate(event: UpgradeTemplateEvent): void {
...
+    templateSet.save();
+}
+
+export function handleEnableFeatureFor(event: EnableFeatureForEvent): void {
+    const feature = getTemplateFeature(
+        event.params.templateId,
+        event.params.featureId,
+    );
+
+    const allowedFeatureAccountId = feature.id.concat(
```
guerrap, line 242: Wondering if we should move this logic to a dedicated function in the `common.ts` file, since we are using this in multiple places.

packages/subgraph/src/kpi-tokens-manager.ts
```diff
@@ -183,3 +222,92 @@ export function handleUpgradeTemplate(event: UpgradeTemplateEvent): void {
...
+    }
+
+    store.remove(
+        "KPITokenTemplateFeatureAllowedAccount",
+        allowedFeatureAccountId.toString(),
+    );
+}
+
+export function handlePauseFeature(event: PauseFeatureEvent): void {
+    const feature = getTemplateFeature(
```
guerrap, line 298: Should we handle the case where the feature being paused doesn't exist?

packages/subgraph/src/kpi-tokens-manager.ts
```diff
@@ -183,3 +222,92 @@ export function handleUpgradeTemplate(event: UpgradeTemplateEvent): void {
...
+        event.params.templateId,
+        event.params.featureId,
+    );
+    feature.paused = true;
+    feature.save();
+}
+
+export function handleUnpauseFeature(event: UnpauseFeatureEvent): void {
+    const feature = getTemplateFeature(
```
guerrap, line 307: Same as above.

packages/subgraph/src/oracles-manager.ts
```diff
@@ -183,3 +218,90 @@ export function handleUpgradeTemplate(event: UpgradeTemplateEvent): void {
...
+    templateSet.save();
+}
+
+export function handleEnableFeatureFor(event: EnableFeatureForEvent): void {
+    const feature = getTemplateFeature(
+        event.params.templateId,
+        event.params.featureId,
+    );
+
+    const allowedFeatureAccountId = feature.id.concat(
```
guerrap, line 238: Same here for the duplicated logic.

packages/subgraph/src/oracles-manager.ts
```diff
@@ -183,3 +218,90 @@ export function handleUpgradeTemplate(event: UpgradeTemplateEvent): void {
...
+    allowed.address = addressToBytes(event.params.account);
+    allowed.save();
+}
+
+export function handleDisableFeatureFor(event: DisableFeatureForEvent): void {
+    const feature = getTemplateFeature(
+        event.params.templateId,
+        event.params.featureId,
+    );
+    const allowedFeatureAccountId = feature.id.concat(
```
guerrap, line 267: Same as above.

packages/subgraph/src/oracles-manager.ts
```diff
@@ -183,3 +218,90 @@ export function handleUpgradeTemplate(event: UpgradeTemplateEvent): void {
...
+    }
+
+    store.remove(
+        "OracleTemplateFeatureAllowedAccount",
+        allowedFeatureAccountId.toString(),
+    );
+}
+
+export function handlePauseFeature(event: PauseFeatureEvent): void {
+    const feature = getTemplateFeature(
```
guerrap, line 292: Same here and below for the existence check.

packages/subgraph/schema.graphql
luzzif: The `featuresOwner` and `features` attributes are actually applied to the template set entities, so `KPITokenTemplateSet` and `OracleTemplateSet`.

packages/subgraph/src/kpi-tokens-manager.ts
luzzif: Oh yes, absolutely. Fixed in 0dedba60ec462b3ca183a04719fe5d88346aa841

packages/subgraph/src/kpi-tokens-manager.ts
luzzif: Updated in 0dedba60ec462b3ca183a04719fe5d88346aa841

packages/subgraph/src/kpi-tokens-manager.ts
luzzif: Addressed in dd931945acc44f5ce2a4e0a39d5590012448dff8

packages/subgraph/src/kpi-tokens-manager.ts
luzzif: In that case, as it is now, it is created and then paused. The contract has no check on this and it happens there too. To be honest I don't think there's any security risk to this, the worst that can happen is that by mistake a certain feature is paused while it shouldn't. At the end of the day a feature is just a mapping from an id to a bool flag.

packages/subgraph/src/kpi-tokens-manager.ts
luzzif: Ditto.

packages/subgraph/src/oracles-manager.ts
luzzif: It should be addressed.

packages/subgraph/src/oracles-manager.ts
luzzif: It should be addressed.

packages/subgraph/src/oracles-manager.ts
luzzif: Ditto.

guerrap (approved):

PR #555, opened 2024-01-04
Add new settings drawer, components refactor

guerrap:


packages/frontend/src/components/chain-select/chain-select.tsx
```diff
@@ -0,0 +1,115 @@
...
+        const handleMouseDown = (event: MouseEvent) => {
+            if (
+                networksPopoverRef.current &&
+                !networksPopoverRef.current.contains(event.target as Node)
+            )
+                setNetworksPopoverOpen(false);
+        };
+        document.addEventListener("mousedown", handleMouseDown);
+        return () => {
+            window.removeEventListener("mousedown", handleMouseDown);
```
luzzif, line 37: I noticed we add the event listener to `document` and then we remove it from `window`. Not 100% sure, but I guess this could result in the listener still being active when the component unmounts.

packages/frontend/src/components/connect-wallet/account-settings-drawer-mobile.tsx
```diff
@@ -0,0 +1,203 @@
...
+        config: {
+            ...springConfig.default,
+            duration: 150,
+        },
+        from: { transform: "translateY(100%)", opacity: 0 },
+        enter: { transform: "translateY(0%)", opacity: 1 },
+        leave: { transform: "translateY(100%)", opacity: 0 },
+    });
+
+    useClickAway(drawerRef, () => {
```
luzzif, line 107: If we want, this is a much less verbose way of doing what we do above through the `useEffect` and the `addEventListener`.

packages/frontend/src/components/connect-wallet/account-settings-drawer-mobile.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const handleDisconnectClick = useCallback(() => {
+        if (!!readonlyConnector)
+            connect({ connector: readonlyConnector, chainId: chain?.id });
+        onClose();
+    }, [connect, onClose, readonlyConnector, chain?.id]);
+
+    const handleSettingsOpenClick = useCallback(() => {
+        setSettingsOpen(true);
+    }, []);
+    const handleSettingsCloseClick = useCallback(() => {
```
luzzif, line 120: Missing a newline here.

packages/frontend/src/components/connect-wallet/index.tsx
```diff
@@ -80,41 +66,25 @@ export const ConnectWallet = ({
...
-        : Error;
     return (
         <>
-            {!__LIBRARY_MODE__ && (
-                <NetworksPopover
-                    open={networksPopoverOpen}
-                    anchor={networksPopoverAnchor}
-                    onNetworkSwitch={handleNetworkSwitchClick}
-                    ref={networksPopoverRef}
+            {width >= 768 ? (
```
luzzif, line 79: Can we transform this into the usage of media queries through Tailwind, which should be more performant?

packages/frontend/src/components/ui/navbar.tsx
```diff
@@ -0,0 +1,141 @@
...
+            const { width } = entries[0].contentRect;
+            if (width > 640) setOpen(false);
+        });
+
+        resizeObserver.observe(document.body);
+        return () => {
+            // eslint-disable-next-line react-hooks/exhaustive-deps
+            resizeObserver.unobserve(document.body);
+        };
+    });
```
luzzif, line 77: I fear a dependency is missing here. Is this hook fired at all without an empty dependencies array?

packages/frontend/src/constants.ts
```diff
@@ -102,26 +103,29 @@ export const CARROT_DOMAIN = prod
...
-        external: true,
-    },
-    {
-        Component: NavLink,
         title: "Campaigns",
+        Icon: Grid,
         to: "/campaigns",
         external: false,
     },
+    {
```
luzzif, line 121: We should probably comment this out for now at least until the new "create" page is ready design wise.

packages/frontend/src/utils/date.ts
```diff
@@ -0,0 +1,16 @@
+import type { Dayjs } from "dayjs";
+import dayjs from "dayjs";
+
+export const getShortDate = (date: Dayjs) => {
```
luzzif, line 4: I'm guessing we could use the less error prone and more readable https://day.js.org/docs/en/durations/humanize, if I understand what this function does.

luzzif (changes requested):

packages/frontend/src/components/connect-wallet/index.tsx
guerrap: The issue here is that with media queries both elements (mobile and desktop) would be attached to the DOM (one with `display: none`), and then both the `useClickAway` would be invoked in either desktop or mobile, causing the drawer to close at any click.

packages/frontend/src/components/chain-select/chain-select.tsx
guerrap: Nice catch, as suggested below I've replaced this with `useClickAway`, fixed in 4b0532268fb3507753b83ec7a80e92c388c81520.

packages/frontend/src/components/connect-wallet/account-settings-drawer-mobile.tsx
guerrap: Fixed in 4b0532268fb3507753b83ec7a80e92c388c81520.

packages/frontend/src/components/ui/navbar.tsx
guerrap: We can actually remove this altogether, fixed in 34e358a7437afad6342fd9a69fb71f1c56d5c4bb.

packages/frontend/src/constants.ts
guerrap: Fixed in 34e358a7437afad6342fd9a69fb71f1c56d5c4bb.

packages/frontend/src/utils/date.ts
guerrap: I think we can use the `relativeTime` plugin, fixed in b83f121041f08ca64b4cbf12a249bb13742c846e.

packages/frontend/src/components/connect-wallet/index.tsx
luzzif: Can in some way `pointer-events: none` help here?

PR #559, opened 2024-01-10
feat(qa): smoke test improvement

MilanVojnovic95:
-

### DIFF
--- a/packages/frontend/src/components/authenticate.tsx
+++ b/packages/frontend/src/components/authenticate.tsx
@@ -81,6 +81,7 @@ export const Authenticate = ({ onCancel }: AuthenticateProps) => {
                     <>
                         <LogoIcon className="w-40 h-40" />
                         <Typography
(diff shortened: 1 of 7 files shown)


.changeset/yellow-berries-allow.md
guerrap, line 1: We can actually delete this file since this change is already in `develop`.

packages/frontend/src/components/authenticate.tsx
```diff
@@ -81,13 +81,15 @@ export const Authenticate = ({ onCancel }: AuthenticateProps) => {
                     <>
                         <LogoIcon className="w-40 h-40" />
                         <Typography
+                            data-testid="welcome-to-carrot-modal-title"
```
guerrap, line 84: This id should follow the convention of `component_name-...`, so for example here it could be `authenticate-modal-title`.

guerrap (approved):

luzzif (approved):

PR #560, opened 2024-01-15
Implement drafts create/load/delete

guerrap:


packages/frontend/src/components/campaign-creation-form.tsx
```diff
@@ -141,39 +163,50 @@ export function CampaignCreationForm<S extends SerializableObject<S>>() {
...
-            }
-            i18n={i18n}
-            className={{ root: "w-full h-full" }}
-            state={draftState.state}
-            onStateChange={handleStateChange}
-            onCreate={handleCreate}
-            navigate={navigate}
-            onTx={addTransaction}
-        />
+        <>
```
luzzif, line 179: Why is there a wrapping React fragment here?

packages/frontend/src/state/reducers/drafts/actions.ts
```diff
@@ -0,0 +1,12 @@
+import type { SerializableObject } from "@carrot-kpi/react";
+import { createAction } from "@reduxjs/toolkit";
+
+export const addDraft = createAction<{
+    draftId: number;
```
luzzif, line 5: Since this is already the draft entity does it make sense for this to just be called `id`?

packages/frontend/src/state/reducers/drafts/actions.ts
```diff
@@ -0,0 +1,12 @@
...
+    templateId: number;
+    draft: SerializableObject<object>;
```
luzzif, line 7: Why `object` here?

packages/frontend/src/components/campaign-creation-form.tsx
guerrap: Fixed in eb18f5629b3ac453997b41920f523ab418f1787e.

packages/frontend/src/state/reducers/drafts/actions.ts
guerrap: Fixed in ec5ed1be03f556cb81c2982b3f68bc703052ae94.

PR #574, opened 2024-01-25
chore: rename data uploader service references to data manager

luzzif:


packages/react/src/hooks/useJSONUploader.ts
```diff
@@ -46,15 +45,15 @@ export const useJSONUploader = <
         async (data: S): Promise<string> => {
             const response = await fetch(
                 `${getServiceURL(
-                    Service.DATA_UPLOADER,
+                    Service.DATA_MANAGER,
                     !devMode && !stagingMode,
                 )}/data/json/s3`,
```
guerrap, line 50: Here should be `/data/s3/json` looking at the service swagger.

PR #590, opened 2024-02-02
refactor: wagmi config

guerrap:


packages/frontend/src/constants.ts
```diff
@@ -33,50 +26,62 @@ export const DATA_CDN_URL = getServiceURL(
...
                   ...gnosis,
-                  logo: GnosisLogo,
-                  iconBackgroundColor: "#04795b",
-                  transport: http(),
-              },
-          ] as const)
-        : ([
+                  rpcUrls: {
+                      default: {
+                          http: ["https://rpc.ankr.com/gnosis"],
```
luzzif, line 36: We can just leave "http()" here. Have a look at the default values for the RPCs. They're the same.

packages/frontend/src/constants.ts
```diff
@@ -33,50 +26,62 @@ export const DATA_CDN_URL = getServiceURL(
...
               {
-                  ...sepolia,
-                  logo: EthereumLogo,
-                  iconBackgroundColor: "#8637ea",
-                  transport: http(),
-              },
+                  ...polygonMumbai,
+                  rpcUrls: {
+                      default: {
+                          http: ["https://polygon-mumbai-pokt.nodies.app/"],
```
luzzif, line 46: Here too.

packages/frontend/src/constants.ts
```diff
@@ -33,50 +26,62 @@ export const DATA_CDN_URL = getServiceURL(
...
-              {
-                  ...polygonMumbai,
-                  logo: PolygonLogo,
-                  iconBackgroundColor: "#8247E5",
-                  transport: http(),
-              },
-          ] as const);
+                  rpcUrls: {
+                      default: {
+                          http: ["https://sepolia-rpc.scroll.io"],
```
luzzif, line 61: And here too.

packages/frontend/src/constants.ts
```diff
@@ -33,50 +26,62 @@ export const DATA_CDN_URL = getServiceURL(
...
+                      },
+                  },
+              } as const satisfies Chain,
+          ];
+
+export const SUPPORTED_CHAIN_ICONS = {
+    [gnosis.id as number]: {
+        logo: GnosisLogo,
+        iconBackgroundColor: "#04795b",
```
luzzif, line 70: We can probably just call this `backgroundColor` or `bgColor`. Also, we should find a way in Typescript to force an error if not all the chain ids from `SUPPORTED_CHAINS` are specified (it would cause a missing icon in the frontend).

packages/frontend/src/library-mode-entrypoint.tsx
```diff
@@ -36,7 +35,9 @@ export const Root = ({
 }: RootProps) => {
     const config = createConfig({
         chains: [supportedChain] as const,
-        transports: { [supportedChain.id]: supportedChain.transport },
+        transports: {
+            [supportedChain.id]: http(supportedChain.rpcUrls.default.http[0]),
```
luzzif, line 39: This is wrong. The RPC URL in this case must be taken directly by the passed in props as we need to reference the local Anvil instance and not the remote chain's RPC.

packages/frontend/src/shared-entrypoint.tsx
```diff
@@ -19,13 +19,12 @@ import { ThemeUpdater } from "./updaters/theme";
...
 import { Fathom } from "./components/fathom";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-import type { WagmiConfig } from "./react-app-env";
 
 dayjs.extend(localizedFormat);
 
 interface SharedEntrypointProps {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    config: WagmiConfig;
+    config: any;
```
luzzif, line 27: Why is it `any` here? Is there no type for a default `wagmi` config?

packages/react/src/hooks/useWagmiPassiveHook.ts
```diff
@@ -16,6 +18,23 @@ export function useWagmiPassiveHook<T>({
 }: UsePassiveHookParams<T>): T {
     const [browserFocused, setBrowserFocused] = useState(true);
 
+    const { data: blockNumber } = useBlockNumber({
```
luzzif, line 21: Shouldn't this only be enabled if the `watch` prop is true? Doesn't this always rerun the hook?

luzzif (changes requested):

packages/react/src/hooks/useWagmiPassiveHook.ts
guerrap: This is enabled based on the condition below:

```ts
enabled: params.query?.enabled && browserFocused
```

It shouldn't run if it's disabled or the browser is not focused.

luzzif: Yes, but this doesn't need to run when the `wagmi` hook is enabled, but when `watch` is true too. Otherwise by default all hooks will be in watch mode.

guerrap: But that's the point of the `useWagmiPassiveHook`, it's a wrapper around the hooks that needs to be in watch mode, but with the browser focused logic added. Maybe I missed your point.

luzzif: Is it only for watch-based hooks?

guerrap: Yes, it's to avoid generating calls for the watch hooks when the browser is not focused.

PR #606, opened 2024-02-28
feat(frontend): limit frontend interaction to same-environment chains

luzzif:


packages/frontend/src/constants.ts
```diff
@@ -23,22 +23,37 @@ export interface CarrotChain extends SupportedChain {
...
-    {
-        ...SUPPORTED_CHAIN[ChainId.ArbitrumSepolia],
-        icon: {
-            logo: ArbitrumLogo,
-            backgroundColor: "#213147",
-        },
-    },
-] as const;
+export const SUPPORTED_CHAINS: [CarrotChain, ...CarrotChain[]] =
+    __ENVIRONMENT__ === "dev"
```
guerrap, line 27: Is this `__ENVIRONMENT__` initialized globally somewhere?

luzzif: It's passed as a global in the craco config. There's also a check that enforces its valid definition.

guerrap (approved):

PR #609, opened 2024-03-06
feat(qa): run tests on diff envs

MilanVojnovic95:
- utility class > giving us ability to choose env to run tests on insided scripts in `package.json`. Example `--env ENV=[env-name]` added for local/dev/stage env for now
- update smoke test to run no dev/stage envs


MilanVojnovic95:
Bots smoke tests are currently passing on DEV & STAGE env - you just have to manually change envs in order to import adequate network `Arbitrum Sepolia` or `Ethereum Sepolia` - both configs are inside `.env.e2e` file 

I will investigate more and try to figure out how to pass configs dinamically. Worst case scenario, we can have 2 different synpress configs containing different networks.

PR #618, opened 2024-03-15
feat: new and improved environments handling

luzzif:


packages/frontend/src/standalone-entrypoint.tsx
```diff
@@ -60,18 +63,19 @@ export const config = createConfig({
...
 export const Root = () => {
-    const setDevMode = useSetDevMode();
+    const setEnvironment = useSetEnvironment();
     const setTemplatePreviewMode = useSetTemplatePreviewMode();
     const setIPFSGatewayURL = useSetIPFSGatewayURL();
 
-    setDevMode(false);
-    setTemplatePreviewMode(__ENVIRONMENT__ !== "prod");
+    setEnvironment(__ENVIRONMENT__);
+    setTemplatePreviewMode(true);
```
guerrap: Should we enable this by default in production too?

packages/frontend/src/standalone-entrypoint.tsx
```diff
@@ -89,7 +93,10 @@ root.render(
     </StrictMode>,
 );
 
-if (__ENVIRONMENT__ === "prod" && "serviceWorker" in navigator) {
+if (
+    __ENVIRONMENT__ !== Environment.Production &&
```
guerrap: Here `__ENVIRONMENT__` should be equal to `Environment.Production`, no?

packages/sdk/src/commons.ts
```diff
@@ -13,36 +13,34 @@ import {
...
-
 export interface ServiceUrls {
-    staticCdn: string;
-    dataCdn: string;
     dataManager: string;
 }
 
 export const SERVICE_URLS: Record<Environment, ServiceUrls> = {
+    [Environment.Local]: {
+        dataManager: "https://data-manager.api.dev.carrot.community",
```
guerrap, line 29: This means that all the campaigns we create locally (when running the playground) will be uploaded to the shared S3 bucket, right? I don't think it's an issue, but it means that local campaign specifications will be uploaded and indexed, and persisted on S3 indefinitely when there's no use for them once the local playground gets turned off; just wanted to bring this point to attention.

packages/frontend/src/standalone-entrypoint.tsx
luzzif: Good point: I've enabled it only in prod here 5e933ddde3dc7906eddd78c3e44602f515b80177

packages/frontend/src/standalone-entrypoint.tsx
luzzif: Good point but I actually opted to only disable the service worker in the development environment, in line with the philosophy of having a staging environment that is as close as possible to production. Addressed in 8283f77c1cde43320b42b180ce930ddd13c721ea

packages/sdk/src/commons.ts
luzzif: You are right. We should probably have a local S3 deployment in `carrot-scripts` alongside the other services (it would replace the current IPFS node start up). Maybe I should add a todo and do this later?

guerrap: Yes, let's add an issue/todo for this, and we can do it later.

guerrap (approved):

packages/sdk/src/commons.ts
luzzif: Issue added: https://github.com/carrot-kpi/monorepo/issues/619

PR #621, opened 2024-03-18
fix: template module load

guerrap:


packages/react/src/hooks/useTemplateModule.ts
```diff
@@ -80,14 +82,24 @@ export const useTemplateModule = <
         )
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             root = template.specification.previewUrl[environment]!;
```
luzzif, line 84: Instead of adding the extra else if block we could probably just add a ternary here and do something like:

```typescript
root = template.specification.previewUrl[environment === Environment.Local ? Environment.Development : environment]!;
```

luzzif (changes requested):

